PR #2, opened 2026-05-27
feat(client): Return `TestResultsJson` from `Client.run`

Sean-Bartie:
## Description:
Allow library consumers to read test results directly after running a client test.

Previously, `Client.run` returned only success or failure, which made it hard for downstream crates to use riperf3 programmatically without extra parsing or CLI workarounds. This change updates Client.run to return Result<TestResultsJson> and surfaces the parsed server results from the control exchange.

If server results are missing, Client.run now returns a protocol error instead of succeeding without data.

## Why this helps:

- Enables clean crate-to-crate integration of riperf3
- Removes the need for external output parsing
- Makes result handling explicit and ergonomic for API users

### DIFF
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -74,7 +74,7 @@ pub struct Client {
 }
 
 impl Client {
-    pub async fn run(&self) -> Result<()> {
+    pub async fn run(&self) -> Result<TestResultsJson> {
         // ---- Generate cookie and connect ----
         let cookie = protocol::make_cookie();
         let mut ctrl = net::tcp_connect(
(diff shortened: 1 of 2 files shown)


therealevanhenry (approved):
Cloned, built, ran the integration suite on Linux — all tests pass. The signature change is API-additive: the only non-test caller (`riperf3-cli/src/main.rs:231`) uses `client.run().await?` and discards the Ok value, so it compiles unchanged. `TestResultsJson` is already reachable via `riperf3::protocol`.

One intentional behavior change worth flagging: the new `ok_or_else` returns a `Protocol` error when `server_results` is `None`, which previously would have been `Ok(())`. That's reachable via the bare `TestState::IperfDone => break` arm (server ends without `ExchangeResults`). The PR description calls this out, and it's the right call — succeeding with no data was the worse outcome.

Added two follow-up tests to lock in the new contract: one for the happy-path return value (non-empty streams, finite cpu_util_total), and one mock-server test that asserts the new `RiperfError::Protocol("missing server results...")` error path.

Issue #60, opened 2026-05-30
Reverse + byte/block limit (-R -n / -R -k) never terminates


therealevanhenry:
## Summary

A **reverse** test with a **byte or block limit** (`-R -n <N>` or `-R -k <N>`) never terminates — the transfer runs indefinitely instead of stopping after `N` bytes/blocks. Duration-limited reverse (`-R -t <s>`) works fine.

## Repro

```
riperf3 -s --one-off -p 5490 &
riperf3 -c 127.0.0.1 -p 5490 -n 512K -R     # hangs; transfers GBytes, never stops
```

Observed: the client (receiver in reverse) keeps printing interval lines well past any reasonable point (11 GByte/interval at 96 Gbit/s on loopback, 20s+), with no termination. `timeout` is required to kill it. The equivalent forward test (`-n 512K`) stops immediately, and `-R -t 1` stops at 1s.

## Likely cause

In reverse, the **server** is the sender and must honor the negotiated byte/block limit (`params.num` / `params.blockcount`) to stop sending — or the client (receiver) must enforce the received-byte limit and drive teardown. One side isn't applying the limit in the reverse direction, so neither stops. (The server's UDP sender path already self-enforces a max *duration* in reverse/bidir for a related reason — see server.rs — but the TCP byte/block limit in reverse isn't wired.)

## Impact

`-R -n` / `-R -k` is a documented iperf3 mode; riperf3 hangs instead of completing — a drop-in incompatibility and a footgun for scripted reverse byte-limited tests.

Found while writing the #50 server-JSON integration test (which initially used `-n` + reverse and hung); the test was switched to duration mode. Pre-existing, unrelated to #50.

Issue #54, opened 2026-05-30
bidir -J intervals: sum lumps both directions; missing sum_bidir_reverse


therealevanhenry:
Surfaced in #36 PR2 review (#53). iperf3 emits three keys per **bidir** interval object — `streams`, `sum`, `sum_bidir_reverse` — splitting the two directions. riperf3's `Interval` model has only `streams` + `sum`, so a `--bidir` run lumps both directions into one `sum` (sender:true) and omits `sum_bidir_reverse`.

Per-stream interval data and the **end** block are correct (the end already emits `sum_*_bidir_reverse`); only the per-interval `sum` is unsplit. Net improvement over main (where `intervals` was empty), but a real parity gap.

Same root as the documented bidir limitation under #36 (riperf3's 2N half-duplex sockets vs iperf3's N full-duplex). Fix: add `sum_bidir_reverse` to the `Interval` model and split the per-interval aggregate by direction, mirroring the end block.

therealevanhenry:
Server-side JSON (#50) validation confirms this gap is **shared by the client and now the server**, and is broader than just `intervals`:

For **bidir UDP** specifically, the `end` aggregates are TCP-shaped (built via the TCP sum path) on both sides:
- `end.sum` is **absent** (iperf3 emits it for single-direction UDP; for bidir UDP iperf3 emits both `sum` and `sum_bidir_reverse`).
- `end.sum_sent` / `sum_received` / `sum_*_bidir_reverse` are **missing the UDP fields** `jitter_ms` / `lost_packets` / `packets` / `lost_percent`.
- the `sender` flags on the bidir-reverse pair are inverted vs iperf3.

Root cause: in `json_report.rs::build()`, the `if self.bidir { ... }` branch precedes `else if is_udp { ... }`, so a **bidir UDP** test never reaches the UDP sum path and gets TCP-shaped aggregates. Verified the riperf3 *client* exhibits the identical gap, so #50 PR1 does not regress it — it's pre-existing here.

Fix should: (1) emit `sum` / `sum_bidir_reverse` for UDP, (2) route bidir UDP through `udp_sum` so the datagram fields are present, (3) correct the bidir-reverse `sender` flags, and (4) add `sum_bidir_reverse` for the interval samples. Because client and server share `build()`, fixing it here corrects both.

Issue #50, opened 2026-05-29
riperf3 server ignores -J/--json-stream (emits text only); blocks server-side interop validation


therealevanhenry:
Surfaced while building the interop gate (#38/#49): riperf3's **server has no JSON output path at all**. `-J` is wired only into the client (`ClientBuilder.json_output` → `print_results_json`); `ServerBuilder`/`Server` have no `json_output` field, and `server.rs` always prints human text (`print_final_summaries`, separator banners). So `riperf3 -s -J` silently prints text, and `--json-stream` is equally ignored server-side (`reporter.rs:430` even hard-codes `json_stream: false` for the server's interval reporter).

### iperf3 behavior
`iperf3 -s -J` emits the test results as JSON on the server's stdout (and `-s -J --json-stream` streams line-delimited JSON).

### Impact
- **Drop-in gap:** anyone scripting a riperf3 *server* with `-J` gets text, not JSON.
- **Blocks server-side interop validation:** the #49 cold review asked the interop gate to validate the *server's* result JSON too (to catch server-side-only regressions, e.g. the #23/#48 teardown class). That was deferred precisely because the riperf3 server emits no JSON — once it does, the gate (`scripts/interop.sh`) should validate the server JSON as well, not just the client's.
- Related to `--get-server-output` (#33), which round-trips the server's output to the client.

### Distinct from #36
#36 is about the *client* `-J` schema faithfulness (empty `intervals`, fabricated `local_*`/`remote_*`, missing fields). **This issue is about the *server* emitting JSON at all.** Both are part of a faithful `-J`; ideally both share the typed-serde-struct model proposed in #36.

### Fix
Give the server a JSON output path gated on `-J` / `--json-stream`, matching iperf3's server JSON, then extend the interop gate to validate the server side of each pairing.

Issue #37, opened 2026-05-29
congestion_used is never populated — applied congestion control isn't reported


therealevanhenry:
Both sides send `congestion_used: None` (`riperf3/src/client.rs:682`, `riperf3/src/server.rs:541`). The congestion-control algorithm that was actually applied is never read back (e.g. via `getsockopt(TCP_CONGESTION)`).

**iperf3 behavior:** reports the CC actually in effect (`congestion_used`), which can differ from what was requested.

**Impact:** affects the results exchange and both `-J` and text output — a user running `-C bbr` gets no confirmation the kernel honored it.

**Fix:** read back `TCP_CONGESTION` after applying it and populate `congestion_used`.

PR #61, opened 2026-05-30
feat(json): server -J / --json-stream output (#50, PR1)

therealevanhenry:
## What

Gives the riperf3 **server** a `-J` / `--json-stream` JSON output path that faithfully matches `iperf3 -s -J`. Previously the server silently ignored both flags and always printed text — a drop-in gap, and the blocker for server-side interop validation (the deferred #49 ask).

Reuses the typed `json_report` model from #36; server-perspective behavior is driven by a new `ReportInput.is_server` flag.

## Server vs client schema (verified empirically vs iperf 3.21)

| aspect | client | server |
|--------|--------|--------|
| connection target | `connecting_to` | **`accepted_connection`** (client's control-socket addr) |
| byte attribution | grafts peer's reported bytes into the un-measured side | **only own measured bytes** — un-measured side is 0 (fwd: sent=0; rev: received=0) |
| `tcp_mss_default` | control-socket MSS | **0** (iperf3's server never reads it); client `-M` still surfaces as `tcp_mss` |
| sender TCP_INFO keys | always emitted | **only on streams the server sent** (fwd receiver's sender block is bytes-only) |
| UDP per-stream | local bytes + peer packets | **sender-side bytes** (0 when received) + measured/derived packets |
| `*_tcp_congestion` | both sides | server's role only (receiver fwd, sender rev/bidir) |
| remote CPU | from server results | **0** (server doesn't surface the client's CPU) |

Listener banners are suppressed under `-J` so stdout is a clean JSON document (matches iperf3).

## Validation

Built a field-by-field structural comparator (riperf3-server vs iperf3-server) across **all six direction×protocol combos**. After iterating to convergence, **every combo is structurally identical** to iperf 3.21. The only remaining diffs are documented gaps **shared with the client** (verified), tracked elsewhere:
- `*_tcp_congestion` absent → **#37** (`congestion_used` not populated yet)
- `snd_wnd`/`max_snd_wnd` = 0 → libc `tcp_info` has no `tcpi_snd_wnd` (documented)
- bidir UDP `sum`/`sum_*` UDP fields → **#54** (the client `build()` bidir branch precedes the UDP branch; same gap on both sides)

Interop gate green vs iperf **3.12 and 3.21** (64/64 each). 9 new model unit tests pin the server attribution rules; an end-to-end integration test exercises the forward + reverse JSON path.

## Scope / follow-ups

- Extend `scripts/interop.sh` to validate the **server** side of each pairing — deferred to **#50 PR2** (now unblocked).
- **#60** (new) — reverse + byte/block limit (`-R -n`) never terminates; pre-existing, found while writing the integration test (which uses duration mode instead).
- `system_info()` moved from `client.rs` to `utils.rs` so client and server share it.

### DIFF
--- a/riperf3-cli/src/main.rs
+++ b/riperf3-cli/src/main.rs
@@ -242,6 +242,12 @@ async fn async_main(cli: Cli) -> std::result::Result<(), Box<dyn std::error::Err
         if cli.verbose {
             builder = builder.verbose(true);
         }
+        if cli.json {
+            builder = builder.json_output(true);
+        }
+        if cli.json_stream {
+            builder = builder.json_stream(true);
+        }
         if cli.daemon {
             builder = builder.daemon(true);
         }
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -886,13 +886,16 @@ impl Client {
                     None
                 };
 
+                // to_canonical(): unwrap an IPv4-mapped IPv6 address to plain IPv4
+                // (matches iperf3); a no-op for the client's usual canonical
+                // addresses, correct if the client is bound to a dual-stack socket.
                 let (local_host, local_port) = s
                     .local_addr
-                    .map(|a| (a.ip().to_string(), a.port()))
+                    .map(|a| (a.ip().to_canonical().to_string(), a.port()))
                     .unwrap_or_else(|| (self.host.clone(), 0));
                 let (remote_host, remote_port) = s
                     .peer_addr
-                    .map(|a| (a.ip().to_string(), a.port()))
+                    .map(|a| (a.ip().to_canonical().to_string(), a.port()))
                     .unwrap_or_else(|| (self.host.clone(), self.port));
 
                 // Sender-side TCP_INFO extremes + real retransmit total collected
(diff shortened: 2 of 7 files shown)


therealevanhenry:
## Review gate satisfied

**Cold review round 1** → BLOCK: two server-JSON fidelity bugs —
1. server bidir interval `sum.retransmits` emitted on the received-flow sum (sender=false) where iperf3 omits it.
2. dual-stack server emitted IPv4-mapped addresses (`::ffff:127.0.0.1`).

Both fixed in dd9d8d3 (gate retransmits on the sum's sender flag; `IpAddr::to_canonical()` on server+client address paths). Comparator filter tightened so a TCP-bidir interval-sum regression is no longer masked.

**Cold review round 2** → **SHIP**: both fixes verified correct and complete vs iperf 3.21 across forward/reverse/bidir × TCP/UDP × single/`-P 4`, and 3.12. Client `-J` confirmed byte-identical to `main` (the `to_canonical` addition is a no-op there). Residual nits are pre-existing client-wide gaps, faithfully inherited, tracked separately: #54 (bidir interval `sum` conflation), #62 (`--json-stream` event envelopes), #57 (float `.0` rendering).

**CI**: 6/6. **Interop**: 64/64 vs iperf 3.12 and 3.21. **Comparator**: all 6 combos structurally identical.

This is PR1 of #50 (server JSON output). PR2 — extending `scripts/interop.sh` to validate the server side of each pairing (now unblocked) — will complete #50, so #50 stays open.

Issue #23, opened 2026-05-29
`PeerDisconnected` Error when running with `--reverse`


Sean-Bartie:
## Summary
When running `riperf3` in reverse mode against `networkstatic/iperf3` in Docker, the transfer runs and interval output is printed for the full test duration, but the client exits with:

`Error: PeerDisconnected`

## Environment
- OS: macOS
- riperf3 version: `0.5.1`
- Server image: `networkstatic/iperf3`
- Server port: `5201` (TCP and UDP mapped)
- Client: local workspace build via `cargo run -p riperf3-cli`

### 1. Start server
```bash
docker run -d --name iperf3-server -p 5201:5201/tcp -p 5201:5201/udp networkstatic/iperf3 -s --verbose -p 5201
```

### 2. Run client (reverse)
```bash
cargo run -p riperf3-cli -- -c 127.0.0.1 --reverse
```

## Actual Behavior
The test exists with`PeerDisconnected` error.

```text
Running `target/debug/riperf3 -c 127.0.0.1 --reverse`
[ ID] Interval           Transfer     Bitrate
[  1]  0.00-1.00  sec  2172.05 MBytes  18220.51 Mbits/sec
[  1]  1.00-2.00  sec  2160.13 MBytes  18120.49 Mbits/sec
[  1]  2.00-3.00  sec  2162.09 MBytes  18136.91 Mbits/sec
[  1]  3.00-4.00  sec  2082.74 MBytes  17471.32 Mbits/sec
[  1]  4.00-5.00  sec  2174.78 MBytes  18243.38 Mbits/sec
[  1]  5.00-6.00  sec  2105.35 MBytes  17660.98 Mbits/sec
[  1]  6.00-7.00  sec  2186.91 MBytes  18345.15 Mbits/sec
[  1]  7.00-8.00  sec  2033.64 MBytes  17059.44 Mbits/sec
[  1]  8.00-9.00  sec  2160.70 MBytes  18125.23 Mbits/sec
[  1]  9.00-10.00 sec  2079.38 MBytes  17443.06 Mbits/sec
Error: PeerDisconnected
```

## Expected Behavior
The reverse-mode test should complete successfully and print final summary output without returning a `PeerDisconnected` error.

## Notes
- This is reproducible with the commands above.
- The data phase appears successful; the failure occurs at/after test completion and control-path teardown.

Sean-Bartie:
This seems to be a compatibility issue with older versions of Iperf3.
```shell
[...]

PR #63, opened 2026-05-30
ci(interop): validate the server's -J JSON too (#50, PR2)

### LINKED ISSUE #50: riperf3 server ignores -J/--json-stream (emits text only); blocks server-side interop validation

therealevanhenry:
## What

Completes **#50** by closing the server-side validation gap the #49 review flagged. The interop harness already ran each server with `-J` and captured its stdout to `$sj`, but only kept it for failure diagnostics — the riperf3 server emitted text, not JSON. Now that the server has a JSON path (#50 PR1, merged), the harness **validates `$sj` on every pairing**.

## Why it matters

A server-side-only regression — the #23/#48 teardown class (connection reset at teardown, zero data, or broken JSON) — can leave the **client** JSON looking fine while the server is broken. Validating the server's own JSON catches that. This is exactly what the #49 cold review asked for and was deferred until the server could emit JSON.

## Details

- **OR, not AND**: the server only populates the one direction it measured (forward → `sum_received`, reverse → `sum_sent`; bidir → both), so `valid_server_result` requires `sum_sent>0 OR sum_received>0` — unlike the client's `AND`. Verified it rejects parse-failure / `error` / zero-transfer and accepts both single-direction server shapes.
- **No truncation**: the one-off (`-s -1`) server now finishes printing and exits on its own before validation (force-killed only if it overruns), so a premature `SIGTERM` can't truncate `$sj` into a false FAIL.

## Validation

Full matrix **64/64 vs iperf 3.12 and 3.21**, no false positives, no retries — locally, across all r→i / i→r / r→r / i→i pairings. Script-only change; the workflow already invokes `scripts/interop.sh` (no workflow edit needed).

Closes #50.

### DIFF
--- a/scripts/interop.sh
+++ b/scripts/interop.sh
@@ -221,9 +221,13 @@ run_case() {
         pass=$((pass + 1))
     else
         printf 'FAIL  %-26s %s\n' "$name" "$col"
-        sed 's/^/    client    | /' "$cj" | tail -4
+        # The validators already print the reason (parse error / no transfer) to
+        # stderr above. These show the raw output for context. The -J files are
+        # pretty JSON whose end-of-test aggregates sit near the bottom, so show
+        # enough of the tail to include the `end` block, not just closing braces.
+        sed 's/^/    client    | /' "$cj" | tail -25
         sed 's/^/    client-err| /' "$cerr" | tail -4
-        sed 's/^/    server    | /' "$sj" | tail -4
+        sed 's/^/    server    | /' "$sj" | tail -25
         sed 's/^/    server-err| /' "$serr" | tail -4
         fail=$((fail + 1))
         failed_cases+=("$name $col")
--- a/scripts/interop.sh
+++ b/scripts/interop.sh
@@ -41,9 +41,9 @@ trap 'rm -rf "$workdir"' EXIT
 # Per-attempt I/O sinks (fixed paths, reused per case). -J JSON (stdout) and
 # diagnostics (stderr) stay SEPARATE: a stray stderr line merged into the JSON
 # sink would make the parser choke and score a good transfer as a false FAIL.
-# The server's output (sj/serr) is kept for failure diagnostics only; it is NOT
-# validated as JSON — the riperf3 server emits text, not JSON yet (#50). Once it
-# gains a -J path, validate sj for the interop pairings too (#49 review).
+# Both sides emit JSON now (the riperf3 server gained a -J path in #50), so the
+# server's output (sj) is validated too — catching server-side-only regressions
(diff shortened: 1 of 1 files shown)


therealevanhenry:
## Review gate satisfied

**Round 1** → SHIP (no blockers): verified `valid_server_result` is non-vacuous (rejects parse-fail/error/zero-transfer, accepts both single-direction server shapes), the kill/wait loop is race-safe, and 9/9 harness runs across iperf 3.12+3.21 showed no flakiness. One nit (FAIL-path `tail -4` only showed closing braces) → fixed (`tail -25`).

**Round 2** → SHIP: independently re-ran the matrix (64/64 ×2, plus a 6-run `INTEROP_DURATION=1` flakiness sweep — all clean), confirmed the 2nd commit is an inert FAIL-only diagnostic, and the OR-check is *required* (an AND would falsely fail every single-direction case). One comment-accuracy nit (bidir aggregates) → fixed.

**CI**: 6/6 on the final head — the iperf 3.12 and 3.21 interop checks run the modified harness in real CI, the definitive validation.

Completes #50.

Issue #34, opened 2026-05-29
-T/--title does not prefix output lines


therealevanhenry:
`-T/--title` is documented as "Prefix output lines" (iperf3 prepends `title:` to every report line). riperf3 sends `title` in the params JSON (`riperf3/src/client.rs:263`) but the reporter never consumes it — no prefix is applied to interval or summary lines (`riperf3/src/reporter.rs`).

**Fix:** thread `title` into the reporter config and prefix each emitted line, matching iperf3.

therealevanhenry:
Fixed in #95 (7d8fb76): client text lines are now prefixed with `<title>:  ` (text mode only; `-J`/`--json-stream` untitled), matching iperf3.

PR #64, opened 2026-05-30
feat(json): emit --extra-data in -J output (#35)

### LINKED ISSUE #35: --extra-data is never emitted in JSON output
`--extra-data` is documented as "Extra data in JSON output". It is sent in params (`riperf3/src/client.rs:264`) but never written to the `-J` output (`print_results_json`, `riperf3/src/client.rs:754-926`).

**iperf3 behavior:** includes `extra_data` in the `start` block of the JSON.

**Fix:** include `extra_data` in the JSON output (see also the `-J` schema-fidelity issue).

therealevanhenry:
## What

`--extra-data` was sent in the parameter exchange but never written to the `-J` output (a silent no-op for JSON). iperf3 emits it as a **top-level** key (after `end`), present only when the flag is given, on **both** the client and the server (the server receives it via params).

riperf3 now matches: `Report.extra_data` (`Option`, `skip_serializing_if`), populated from the client's `--extra-data` and from `params.extra_data` on the server.

## Validation (vs iperf 3.21)

- `--extra-data X` → top-level `"extra_data": "X"` (not nested in `start`), on client and server; absent without the flag. Confirmed both sides match iperf3's placement.
- Server-JSON comparator: still structurally identical across all 6 direction×protocol combos.
- Interop gate: 64/64 vs iperf 3.12 and 3.21. 2 new model unit tests.

Closes #35.

### DIFF
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -991,6 +991,7 @@ impl Client {
             gso: i32::from(self.gsro),
             gro: i32::from(self.gsro),
             start_time_millis: start_meta.start_time_millis,
+            extra_data: self.extra_data.clone(),
             intervals: collected_intervals,
             streams: stream_reports,
         };
--- a/riperf3/src/json_report.rs
+++ b/riperf3/src/json_report.rs
@@ -36,6 +36,11 @@ pub struct Report {
     pub start: Start,
     pub intervals: Vec<Interval>,
     pub end: End,
+    /// `--extra-data` string, emitted at the top level (after `end`) only when
+    /// given — matching iperf3's placement (#35). Present on both client and
+    /// server (the server receives it via the parameter exchange).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub extra_data: Option<String>,
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


therealevanhenry:
## Review gate satisfied

**Round 1** → SHIP: placement (top-level, after `end`), presence (only with the flag), client (`self.extra_data`) + server (`params.extra_data`) sourcing, and value/escaping all verified byte-identical to iperf 3.21. Pre-existing CLI divergence (server accepts+ignores client-only `--extra-data` where iperf3 rejects it) filed as #65, out of scope here.

**Round 2** → SHIP: independently re-verified, incl. edge cases — empty string (`--extra-data ""` → `"extra_data": ""` on both), 4KB string, `"`/`\`/tab/newline escaping, raw UTF-8. `cargo test` 177+129+59, clippy clean, comparator 0 unexpected diffs.

**Note for the future #34 (title):** iperf3's top-level finish order is `title, extra_data, server_output_*`. riperf3 emits only `extra_data` today, so after-`end` is faithful — but when `title` is added it must serialize *before* `extra_data` to keep iperf3's key order.

CI 6/6; interop 64/64 vs iperf 3.12 and 3.21.

Issue #40, opened 2026-05-29
macOS advertises retransmit info but always reports 0


therealevanhenry:
`has_retransmit_info()` returns `true` on macOS (`riperf3/src/tcp_info.rs:131`), but the macOS `get_tcp_info` hardcodes `total_retransmits: 0` (`riperf3/src/tcp_info.rs:81`) because macOS's `tcp_connection_info` only exposes retransmitted *bytes*, not a packet count.

**Impact:** on macOS the interval report shows a `Retr` column that is always `0`, implying a loss-free transfer regardless of reality.

**Fix:** either remove macOS from `has_retransmit_info()`, or suppress the `Retr` column when no usable count is available, rather than printing a misleading 0.

therealevanhenry:
Addressed in #95 (7d8fb76): macOS no longer prints a misleading `Retr 0` (it stops advertising retransmit info; the coupled Cwnd column is suppressed with it). The *faithful* fix — showing real macOS Retr/Cwnd via `tcpi_txretransmitpackets` — is tracked in #96.

Issue #39, opened 2026-05-29
CI only cargo-checks macOS/Windows; FreeBSD has no CI


therealevanhenry:
`.github/workflows/ci.yml` builds + tests **Linux only**. The `cross-check` job runs `cargo check` (no codegen, link, or run) for `x86_64-pc-windows-msvc` and `aarch64-apple-darwin`, and **FreeBSD is not in the matrix at all**.

Consequences:
- "compiles" is unproven for macOS/Windows (`cargo check` skips codegen/link).
- "runs" is unproven for every non-Linux target.
- FreeBSD-specific code is compiled by nobody (e.g. the FreeBSD `sendfile` arity at `riperf3/src/stream.rs:448`, the FreeBSD `tcp_info` fields at `riperf3/src/tcp_info.rs:115-122`).

**Fix:** at minimum `cargo build` + `cargo test` on macOS and Windows runners; add FreeBSD via a cross/VM job (e.g. `cross`, or a `vmactions/freebsd-vm` step).

therealevanhenry:
Resolved: native CI now covers all three non-Linux platforms. macOS + Windows native build+test landed in 0.6.0; FreeBSD (vmactions/freebsd-vm, build+test) landed in 0.6.1 and is promoted to a required check. macOS/Windows native jobs remain informational (Windows tracks #80).

PR #69, opened 2026-05-30
ci: native macOS/Windows test runners + musl check (#39, informational)

therealevanhenry:
## What

Adds three CI jobs (all **informational** — not in branch protection's required checks):
- **`Native test (macos-latest)`** + **`Native test (windows-latest)`** — full `cargo test --workspace` on native runners.
- **`musl check`** — `cargo check --workspace --target x86_64-unknown-linux-musl`.

## Why

The existing `Cross-platform (…)` jobs only `cargo check` the Windows/macOS targets, **cross-compiled from Ubuntu** — they catch compile errors but never build or run natively, so platform-specific *runtime* behavior is unverified. riperf3 has genuinely divergent platform code (the macOS `tcp_connection_info` path in `tcp_info.rs`, winsock) — e.g. **#40** (macOS retransmits always 0) is exactly the class of bug a compile-check can't see. musl matters because riperf3 ships in containers, where libc/`nix` syscall differences surface.

Mainstream Linux is already the best-covered platform (Ubuntu runs the full suite + lint + both interop matrices), so this fills the actual gap: native macOS/Windows execution.

## Informational-first (per plan)

These jobs are **not required-to-merge**. Turning native tests on may surface pre-existing platform bugs, so the plan is: **run informationally → file issues for what reds → fix → then promote to required.** The existing `cross-check` stays the fast required compile gate.

I'll triage whatever the native jobs surface on this PR and file issues before merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,3 +62,47 @@ jobs:
         with:
           key: ${{ matrix.target }}
       - run: cargo check --target ${{ matrix.target }}
+
+  # Native build + test on macOS and Windows (#39). The cross-check job above
+  # only cargo-checks those targets (cross-compiled from Linux) — it never builds
+  # or runs them natively, so platform-specific runtime behavior (e.g. the macOS
+  # tcp_connection_info path, winsock) is unverified. These jobs run the real test
+  # suite on native runners.
+  #
+  # INFORMATIONAL for now: they are deliberately NOT in branch protection's
+  # required checks, because turning native tests on may surface pre-existing
+  # platform bugs (e.g. #40, macOS retransmits always 0). They report without
+  # blocking merges; issues get filed + fixed, then the checks are promoted to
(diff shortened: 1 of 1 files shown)


therealevanhenry:
Native jobs ran and surfaced exactly what informational-first is meant to catch — two test-portability issues, now filed:
- **#71** — Windows: `tcp_info` test uses Unix-only `as_raw_fd()` → build fails. (`musl check` passed, so the lib compiles for musl.)
- **#72** — macOS: `bind_device_loopback` hardcodes `"lo"` + `--bind-dev`/`SO_BINDTODEVICE` is Linux-only → 1 test fails (122 pass).

All 7 required checks pass; the native jobs are intentionally non-required, so this is mergeable. Plan: merge → fix #71/#72 (small `#[cfg]` gates) so the native jobs go green → promote `Native test (…)` to required.

Issue #79, opened 2026-05-30
Windows: --cport (client port binding) fails with WSAEWOULDBLOCK (non-blocking connect not awaited)


therealevanhenry:
## Summary

On `windows-latest` (native CI, #39), `implemented_flag_tests::client_port_binding` (integration.rs:1806) fails:

```
--cport failed: Err(Io(Os { code: 10035, kind: WouldBlock,
  message: "A non-blocking socket operation could not be completed immediately." }))
```

`10035` = `WSAEWOULDBLOCK`. When the client binds a specific source port (`--cport`) and connects, the non-blocking `connect()` on Windows returns `WSAEWOULDBLOCK` (connect in progress) instead of completing or blocking. The code path apparently treats that as a hard error rather than awaiting writability (connect completion).

## Why it's a real bug (not a gate)

`--cport` is a cross-platform feature (iperf3 supports `--cport` everywhere). The 116 other Windows tests pass, so the general transfer path works — this is specific to the bind-cport connect path on Windows. Gating would hide a genuine Windows gap.

## Likely fix

In the connect path used for `--cport`, handle `WSAEWOULDBLOCK`/`WouldBlock` on Windows by awaiting socket writability (poll/select for connect completion) rather than erroring — tokio's connect should normally handle this, so the issue may be in a manual bind+connect path for the explicit source port. Investigate `net.rs`/the cport bind-then-connect flow on Windows.

Deferred per the 0.6.0 plan (native CI stays informational until Windows bugs land). Tracked from #76 (part B).

PR #85, opened 2026-05-30
fix(udp): Windows UDP --bidir -P 4 hangs (#80)

### LINKED ISSUE #80: Windows: UDP --bidir -P 4 hangs (native test timeout)
## Summary

On `windows-latest` (native CI, #39), `implemented_flag_tests::udp_bidir_parallel_completes` (integration.rs:1964) fails — the test's own message: **"UDP --bidir -P 4 hung — issue #5 regression"**. So UDP in bidirectional mode with 4 parallel streams hangs on Windows (the test times out waiting for completion).

## Why it's a real bug (not a gate)

UDP bidir + parallel is cross-platform. Single-direction UDP and other parallel tests pass on Windows (116 pass), so this is specific to the UDP-bidir-parallel path on Windows — likely a start-barrier / stream-setup ordering issue (the test references the #5 class: UDP senders parked on the start barrier). Windows UDP socket semantics (connected UDP, WSAEWOULDBLOCK on the parallel handshake) may interact with the barrier/setup differently than on Unix.

## Investigation

Reproduce on a Windows runner; check the UDP create-streams handshake + start-barrier path (server.rs / stream.rs UDP setup) under `-u --bidir -P 4` on Windows. Compare to the Unix path. May share a root cause with the `--cport` WSAEWOULDBLOCK issue (Windows non-blocking socket handling).

Deferred per the 0.6.0 plan (native CI informational until Windows bugs land). Tracked from #76 (part B).

therealevanhenry:
Fixes #80.

## Root cause

The integration test fails with the "hung" assertion (a 20s timeout), not an error — so setup isn't the problem (the UDP connect handshake is fully bounded and would *error* on failure). The hang is the client's teardown join: `for s in streams { s.task.await }` waiting on a blocking UDP **sender** thread that never returns.

On non-Unix, `net::configure_udp_sender` only switched the socket to blocking — **no send timeout** (the old comment literally said "no portable SO_SNDTIMEO here"). A blocking `send()` whose buffer stays full blocks indefinitely. `udp --bidir -P 4` at 50 Gbps/stream saturates the loopback buffers; at teardown the peer's receivers stop draining, so an in-flight `send()` on the client's sender wedges. The per-batch `done`/deadline checks only run *between* sends, so the thread never observes the stop. On Unix this can't happen — `configure_udp_sender` sets `SO_SNDTIMEO` (1s), which breaks the wedge.

Windows confirmed on this branch's parent run: `client_port_binding ... ok`, only `udp_bidir_parallel_completes ... FAILED` (panic at integration.rs:1971, the "hung" assertion).

## Fix

- Give the non-Unix `configure_udp_sender` a write timeout via `std::net::UdpSocket::set_write_timeout` — portable, maps to `SO_SNDTIMEO`. `UDP_SEND_TIMEOUT_MS` is now defined for all platforms (was `#[cfg(unix)]`).
- `run_udp_sender_blocking` now treats `TimedOut` like `WouldBlock`: a zero-progress batch that rewinds the sequence and re-checks `done`/deadline. (Previously `TimedOut` hit the catch-all arm, which also broke the batch but logged + slept per occurrence.)

No effect in the common case (a healthy send completes immediately; the timeout only fires on a wedge). Unix behavior is unchanged.

## Tests (TDD)

- Pre-existing red end-to-end test: `udp_bidir_parallel_completes` on `windows-latest` (can't run on the Linux dev box — verified via this PR's `Native test (windows-latest)` job).
- Added `net::tests::configure_udp_sender_sets_write_timeout` (`#[cfg(not(unix))]`), the Windows analog of the existing Unix `SO_SNDTIMEO` test — unit-level red→green.

## Local verification (Linux)
- `cargo test --workspace`, `cargo clippy --all-targets -D warnings`, `cargo fmt --check` — clean
- `cargo check --target x86_64-pc-windows-msvc --tests` — non-Unix path + new test compile

Once this and #79 are on `main`, `Native test (windows-latest)` goes green and gets promoted to a required check (separate step).

### DIFF
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -171,6 +171,7 @@ impl Client {
         // ---- State machine: react to server-driven transitions ----
         loop {
             let state = protocol::recv_state(&mut ctrl).await?;
+            eprintln!("[riperf3 #80 trace] client state -> {state:?}"); // TEMP DEBUG
 
             match state {
                 TestState::ParamExchange => {
@@ -241,10 +242,22 @@ impl Client {
         }
 
         // ---- Clean up ----
+        eprintln!(
+            "[riperf3 #80 trace] client cleanup: setting done, awaiting {} tasks",
+            streams.len()
+        ); // TEMP DEBUG
         done.store(true, Ordering::Relaxed);
         tokio::time::sleep(Duration::from_millis(100)).await;
         for s in streams {
+            eprintln!(
+                "[riperf3 #80 trace] awaiting task id={} sender={}",
+                s.id, s.is_sender
+            ); // TEMP DEBUG
             let _ = s.task.await;
+            eprintln!(
+                "[riperf3 #80 trace] joined  task id={} sender={}",
+                s.id, s.is_sender
+            ); // TEMP DEBUG
         }
 
         server_results.ok_or_else(|| {
@@ -477,7 +490,9 @@ impl Client {
                         net::set_bind_dev(&udp_sock, dev)?;
                     }
                     udp_sock.connect(remote).await?;
+                    eprintln!("[riperf3 #80 trace] udp stream {i}/{total}: handshake start"); // TEMP DEBUG
                     protocol::udp_connect_client(&udp_sock).await?;
+                    eprintln!("[riperf3 #80 trace] udp stream {i}/{total}: handshake ok"); // TEMP DEBUG
 
                     // Apply GSO/GRO if requested (no-ops on non-Linux)
                     if self.gsro {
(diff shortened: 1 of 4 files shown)


therealevanhenry:
Closing: the write-timeout change here does not fix #80. Root cause is a native-winsock UDP multi-socket demux limitation (full analysis posted on #80), which needs a server UDP-accept rewrite (userspace demux) — out of scope for a tight 0.6.1. Per the scope decision, the Windows native check stays informational this release and #80 stays open. The #79 cport fix ships separately (#84).

PR #148, opened 2026-06-09
fix(net): apply --bind-dev before connect() for TCP (#88)

### LINKED ISSUE #88: --bind-dev is applied after connect() for TCP sockets (may not constrain routing)
Surfaced during the #72 cold review.

For TCP, `set_bind_dev` is called **after** the socket is already connected:
- control socket: `client.rs` connects in `tcp_connect`, then `set_bind_dev(&ctrl, dev)` runs afterward;
- data sockets: connected in `create_streams`, then `set_bind_dev` runs afterward.

`SO_BINDTODEVICE` (Linux) / `IP_BOUND_IF` (macOS) affect routing/egress interface selection, which is decided at connect time. Setting them on an already-established TCP connection generally does not re-route it, so `--bind-dev` may be a silent no-op for TCP at runtime. The UDP path binds before connect (`create_streams` UDP branch), which is correct.

Tests can't catch this because they run loopback-to-loopback, where bound-vs-unbound is indistinguishable. The `bind_device_invalid_rejects` negative test still proves `set_bind_dev` is *invoked* and its error path is wired; it just can't prove routing takes effect.

Fix direction: apply `set_bind_dev` to the socket **before** `connect()` (build the socket via socket2, set the device, then connect) for the TCP control + data paths, mirroring the UDP ordering.

Related: macOS `set_bind_dev` only handles IPv4 (`IP_BOUND_IF`); an IPv6 socket needs `IPV6_BOUND_IF`. macOS `--bind-dev` over IPv6 is currently unverified/likely-ineffective and should be addressed in the same change.

Not in 0.6.1 scope (compatibility/CI release); filing to track.

therealevanhenry:
## Summary

`SO_BINDTODEVICE` (Linux) / `IP_BOUND_IF` (macOS) affect the routing/egress decision made at connect time — setting them on an already-established TCP connection generally does not re-route it, so `--bind-dev` was a silent no-op for TCP at runtime (control + data sockets both applied it post-connect). The UDP path already bound before connect and is unchanged (it only picks up the new address-family argument).

- `net::tcp_connect` gains a `bind_dev` parameter and applies `set_bind_dev` to the **unconnected** socket2 socket; `bind_dev` joins `--cport`/`-B`/mptcp as a trigger for the socket2 path.
- The post-connect `set_bind_dev` calls on the control socket and TCP data sockets are removed.
- macOS IPv6: `IP_BOUND_IF` only affects the v4 stack; `set_bind_dev` now takes the address family and uses `IPV6_BOUND_IF` at `IPPROTO_IPV6` for v6 sockets (#88's companion note).

## TDD

New test `bind_device_applied_before_connect` pins the ordering **behaviorally**: with no server listening, an invalid device must surface the device error (raised pre-connect), not connection-refused (the post-connect symptom). Confirmed red before the fix (failed with `Connection refused (os error 111)`), green after.

`bind_device_invalid_rejects` needed a scaffolding fix: post-#88 the client errors before ever connecting, so its one-off server never sees a connection — awaiting it hangs forever; it is now aborted.

## Verification limits

Loopback CI cannot prove the routing effect (bound-vs-unbound is indistinguishable on lo) — that's exactly why this bug survived. The behavioral ordering test above is the strongest in-repo signal; real-interface verification (bind to the correct/wrong device on the sandbox fleet and observe connectivity) is queued via riperf3-verify before the 0.7.2 release.

Closes #88.

### DIFF
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -144,6 +144,7 @@ impl Client {
             self.connect_timeout,
             None,
             self.bind_address.as_deref(),
+            self.bind_dev.as_deref(),
             self.mptcp,
             self.ip_version,
         )
@@ -165,10 +166,8 @@ impl Client {
             self.blksize
         };
 
-        // Apply control connection options
-        if let Some(ref dev) = self.bind_dev {
-            net::set_bind_dev(&ctrl, dev)?;
-        }
+        // Apply control connection options (bind_dev is applied inside
+        // tcp_connect, pre-connect — #88)
         if let Some(ref spec) = self.cntl_ka {
             let (idle, intv, cnt) = parse_keepalive(spec);
             net::set_tcp_keepalive(&ctrl, idle, intv, cnt)?;
@@ -418,6 +417,7 @@ impl Client {
                         self.connect_timeout,
                         self.cport,
                         self.bind_address.as_deref(),
+                        self.bind_dev.as_deref(),
                         self.mptcp,
                         self.ip_version,
                     )
@@ -445,9 +445,6 @@ impl Client {
                     }
                     if let Some(label) = self.flowlabel {
                         net::set_ipv6_flowlabel(&data_stream, label)?;
-                    }
-                    if let Some(ref dev) = self.bind_dev {
-                        net::set_bind_dev(&data_stream, dev)?;
                     }
                     if self.tos != 0 {
                         net::set_tos(&data_stream, self.tos as u32)?;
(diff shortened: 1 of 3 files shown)


therealevanhenry:
Post-merge verification on the sandbox fleet (real second host, real NICs — the loopback-CI blind spot this bug lived in):

- `--bind-dev enp0s2` (correct device) → full run, 74.8 Gbit/s.
- `--bind-dev lo` (wrong device) → `ConnectionTimeout`, exit 1 — the binding now genuinely constrains routing; pre-fix this completed normally because the post-connect option was a no-op.
- `--bind-dev nodev0` → pre-connect `No such device`, never connects.

Unprivileged `SO_BINDTODEVICE` confirmed working on the fleet's 6.12 kernel. Post-merge smoke: compat 48/48 PASS, N=1 bench drift-free vs the 0.7.1 baseline.

Issue #124, opened 2026-06-08
CLI wiring tests duplicate main.rs builder logic + miss some flags


therealevanhenry:
Surfaced during #123 cold review (pre-existing on `main`, deferred to keep #123 scoped to encapsulation).

## Duplication
`cli_wiring_tests::build_client_from_cli` (`riperf3-cli/src/cli.rs`) is a hand-maintained **copy** of `main.rs`'s `Cli → ClientBuilder` wiring. The tests therefore prove "helper == expected", not "main.rs == expected". If `main.rs` and the helper drift (a flag wired in `main.rs` only), no test catches it — the same blind spot the `-J` bug exploited, just moved one level up.

**Fix:** extract the `Cli → ClientBuilder` (and `Cli → ServerBuilder`) mapping into one production function in `riperf3-cli` that both `main.rs` and the tests call. (This was the "+dedup" option deliberately deferred from #123.)

## Coverage gaps (same root cause)
Flags wired in `main.rs` but with no dedicated `_wired` test: `port`, `gsro`, `zerocopy`, `interval`, `cntl_ka`. The test helper itself also omits some flags `main.rs` sets (`format`, `username`, `rsa_public_key_path`, `use_pkcs1_padding`) — so even a whole-config compare can't catch their miswiring. Extracting the shared mapping (above) makes these naturally covered.

Non-breaking, CLI-test-only; can land anytime.

Issue #65, opened 2026-05-30
Client-only flags (e.g. --extra-data) accepted + silently ignored on the server; iperf3 rejects them


therealevanhenry:
## Summary

iperf3 marks several options **client-only** and **hard-errors** when they're passed to the server. Example:

```
$ iperf3 -s --extra-data X
iperf3: parameter error - some option you are trying to set is client only
```

(enforced via `client_flag=1` on the option + the `role=='s' && client_flag` gate in `iperf_api.c`.)

riperf3 instead **accepts** these flags on the server and **silently ignores** them. `riperf3 -s --extra-data X` runs cleanly and drops the value. This is a drop-in divergence: a script that relies on iperf3 rejecting a misplaced client-only flag gets different behavior.

## Scope

`--extra-data` is the concrete case found during #35 (it's a global clap arg consumed only in the client branch of `riperf3-cli/src/main.rs`; the server branch never reads it). The same pattern likely affects other client-only options that riperf3 exposes globally (e.g. `-R/--reverse`, `--bidir`, `-b/--bitrate`, `-t/--time`, `-T/--title`, `--fq-rate`, `-M`, `-l`, `-n`, `-k`, `-O`, ...). Worth auditing the full client-only set against iperf3's option table.

## Fix options

1. **Match iperf3**: reject client-only options when running as a server (`-s`), with iperf3's "client only" error and a non-zero exit. Faithful to iperf3's behavior.
2. Looser: warn-and-ignore (less faithful).

The manifesto favors matching iperf3, so option 1 — but it's a behavioral change across many flags, so it deserves its own focused PR with the option set audited against iperf3. Pre-existing (the flags were already global); surfaced during #35 cold review.

PR #153, opened 2026-06-09
fix: reject conflicting end conditions (#140) + consistent fatal TOS policy (#45)

### LINKED ISSUE #45: Inconsistent error handling for socket options (? vs let _ =)
The same socket option propagates errors on one path and swallows them on another. Example: TCP applies `net::set_tos(&data_stream, ...)?` (`riperf3/src/client.rs:369`) while the UDP path uses `let _ = net::set_tos(&udp_sock, ...)` (`riperf3/src/client.rs:454`). Similar `let _ =` vs `?` inconsistencies appear across client/server option application.

**Fix:** decide per option whether a failure is fatal or best-effort, document the choice, and apply it consistently across TCP/UDP and client/server.

### LINKED ISSUE #140: -t accepted together with -n/-k; iperf3 rejects (IEENDCONDITIONS)
Follow-up from the 0.7.1 review.

iperf3 permits only one test end condition: giving `-t` together with `-n` or `-k` fails with `IEENDCONDITIONS` — "only one test duration option may be set (-t, -n or -k)". riperf3 accepts the combination silently (the byte/block end condition wins and `-t` is ignored), so a script relying on iperf3's rejection gets a silently different test instead of an error.

**Fix:** reject the combination with iperf3's wording, following the #100 pattern (client-side rejection of invalid flag combinations, exercised through the real `Cli::build_client` mapping per #124). Decide the layer to match where iperf3 errors (parameter validation, not clap parse), so the exit code/message shape stays faithful.

**Scope note:** also check `-n` + `-k` together (iperf3 rejects that pairing under the same error) and add wiring tests for all three pairings.

therealevanhenry:
## Summary

Two small faithfulness/consistency items, one PR (both touch option validation/error semantics):

**#140 — conflicting end conditions now rejected.** iperf3's `parse_arguments` checks pairwise over `duration_flag`/`bytes`/`blocks` and fails with `IEENDCONDITIONS`; riperf3 silently let `-n`/`-k` win over `-t`, so scripts relying on iperf3's rejection ran a different test instead of erroring. The check lives in `Cli::build_client` — the CLI layer is where flag *explicitness* is knowable (iperf3's check is on explicit flags; a default duration never conflicts), and per #124 the wiring tests exercise the real mapping. Message matches iperf3's verbatim: `only one test end condition (-t, -n, -k) may be specified`.

**#45 — `set_tos` policy made consistent (fatal everywhere).** The same option propagated with `?` on the client TCP path but was swallowed with `let _` on the client UDP and server TCP paths. iperf3's `iperf_common_sockopts` makes TOS failure fatal (`IESETTOS`/`IESETCOS`) on both roles and both protocols, so all four riperf3 sites now use `?`. GSO/GRO keeps `let _` deliberately — riperf3-exclusive optimization, degrade-don't-fail — now documented at the site. The `let _` sites inside net.rs helpers (e.g. buffer-size fallbacks) are separately-documented single-path behaviors, not cross-path inconsistencies, and are out of scope.

## TDD

`conflicting_end_conditions_rejected` (all three pairings) was red against the old `build_client`, green after; `single_end_conditions_accepted` pins the non-conflicting cases. The #45 failure paths aren't triggerable on loopback CI (unprivileged `IP_TOS` accepts any u8), so that half is policy-by-inspection against the iperf3 source — the full suite pins that success paths are unchanged (438 tests green).

## Follow-up noticed (filed separately)

The server's UDP path never applies TOS at all (iperf3 applies `iperf_common_sockopts` to UDP streams on both roles) — filed as its own issue since the UDP demux architecture makes it non-trivial.

Closes #140. Closes #45.


therealevanhenry:
Round-1 findings addressed in 0235262:

1. **[BLOCKER] value-based check** — `end_conditions_conflict()` now mirrors `parse_arguments` exactly: only `-t` is flag-based; the `-n`/`-k` legs test the parsed value (`-t 5 -n 0` accepted, runs the duration test). New red→green tests cover all three zero pairings. The builder additionally normalizes `bytes/blocks 0 → unset`, fixing the adjacent pre-existing divergence (`-n 0` alone used to end instantly; iperf3 runs a duration test). Unparseable `-n abc` deliberately surfaces the parse error (documented riperf3 improvement), not a bogus conflict.
2. **Placement** — the check now ALSO runs in `main` before any side effects (#65/#100 pattern); process-level test pins exit 1 + stderr text + **no pidfile created**. `build_client` keeps the same check for the wiring-test path (#124).
3. **GSO/GRO comment** — corrected: iperf3 3.20+ has `--gsro` with the same disable-and-continue policy; the rationale now cites `iperf_udp_gso/gro`.
4. **set_tos family-aware** — v6 sockets now set `IPV6_TCLASS` (fatal, IESETCOS) + tolerated `IP_TOS` for the mapped-v4 case, per `iperf_common_sockopts`; v4 keeps fatal `IP_TOS`. Removes the BSD risk of the newly-fatal sites AND makes `-S` actually mark native-IPv6 traffic (round 1 demonstrated it silently didn't).
5. **Process-level exit-code test** — added (`tests/end_conditions.rs`).

therealevanhenry:
Round-2 blockers fixed: `set_tclass_v6` (socket2, feature already on) replaces the Linux-only nix sockopt — FreeBSD/macOS/NetBSD/Windows checks all compile again; `end_condition_set` now scale-then-truncates like `unit_atoi` (`-n 0.5` accepted, runs duration). Nits: build_client rustdoc re-attached, setter 0=unset docs added, stale parity comment corrected. On Q6 (server-UDP TOS): tracked as #154 — #45 closes with this PR, the UDP-arm application gap stays open there.

therealevanhenry:
Round-3 item addressed verbatim: the two prescribed rows added (`-t 5 -n 0.5` → accepted set; `-t 5 -n 0.5K` → rejected set) — they distinguish round-2 from round-3 semantics, so a regression to strip-suffix-test-nonzero now fails the suite. The incidental `-S` hex/octal finding is filed separately. Three substantive cold rounds complete; merging on green CI per the reviewer's rubber-stamp note.

Issue #103, opened 2026-06-05
-n/-k byte/block-limited summary uses default duration window (wrong secs + bitrate)


therealevanhenry:
Surfaced during #101 review (pre-existing).

A byte-limited run (`-c <host> -n 25G`) that finishes at ~2.43s prints a summary line `0.00-10.00 sec ... 25958 Mbits/sec` — it uses the default `self.duration` (10) for the summary window and rate instead of the actual elapsed time. The per-interval final partial (`2.00-2.43`) is correct (#55); only the **summary** line is wrong.

Fix: the client summary path should use the measured elapsed test time (not `self.duration`) when the end condition is byte/block-limited. Check `build_results`/`print_results` in client.rs (the `test_duration = self.duration` sites the #55 review flagged at client.rs:743/816/891).

Issue #55, opened 2026-05-30
Final interval occasionally dropped from output (text + -J)


therealevanhenry:
Surfaced in #36 PR2 review (#53); **pre-existing on main**, not introduced by that PR. The interval reporter loop is `tick -> if done break`, so the in-progress final interval is skipped if `done` is set before its tick lands. A 2s `-i 1` run intermittently yields 1 interval instead of 2 (text, json-stream, and now the `-J` intervals array alike). iperf3 reliably emits the final (partial) interval.

Fix: on teardown, flush a final partial interval for the elapsed remainder before exiting the reporter loop (compute the last tick's bytes/stats up to `done`).

Issue #147, opened 2026-06-09
Client error paths leak the interval-reporter task (run_test ServerTerminate returns before finish/done)


therealevanhenry:
Verified during the 2026-06-09 architecture review follow-up. (The review's original hypothesis — a finish/done double-report race — is **wrong**: the reporter's `biased` select and the #55 boundary-fold handle that correctly. The real defect is adjacent.)

## The leak

`Client::run_test` spawns the interval reporter (client.rs:706) and stops it via `reporter_end.finish(end_secs)` + `done.store(true)` + handle await at the end (client.rs:790-794). But the duration-mode select arm returns early on `ServerTerminate`:

```rust
state = protocol::recv_state(ctrl) => {
    if let Ok(TestState::ServerTerminate) = state {
        return Err(RiperfError::Aborted("server terminated".into()));  // client.rs:758
    }
}
```

That path exits with the reporter task detached: `done` is never set and `finish` never fires, so the task ticks forever. Invisible in the CLI (the process exits), but a **library consumer** with a long-lived tokio runtime leaks one reporter task per aborted run — printing zero-byte interval lines every `-i` seconds indefinitely when interval printing is on.

The outer `run()` state loop has the same shape in miniature: any `?` in the loop (send/recv results, send_state) skips the `done.store(true)` + stream-task drain at client.rs:295-299 — the data-stream tasks are mostly self-limiting (sender deadlines, peer-close on receive), but the cleanup contract is happy-path-only. Server side worth auditing for the same pattern (server.rs:569-570).

## Fix direction

Make reporter shutdown unconditional in `run_test`: capture the end-condition result instead of early-returning, run `finish`/`done`/handle-await, then propagate (`let r = ...; reporter_end.finish(..); done.store(..); let _ = handle.await; r?`). Equivalent guard for the outer loop's `done`.

**Test (observable without timing games):** with a collector attached, `Arc::strong_count` on the collector returns to 1 after `run()` errors — proving the reporter task ended; plus a mock server that sends `ServerTerminate` mid-test.

0.7.2-eligible: small, non-breaking, and the lib-consumer impact is real (#137's audience).

therealevanhenry:
Fixed in 0.7.2 by PR #166: the abort path joins the reporter (client.rs ~893-903), pinned by the Arc-refcount leak test.

PR #162, opened 2026-06-09
fix(report): real -O/--omit semantics (#31)

### LINKED ISSUE #31: -O/--omit is cosmetic: warm-up not excluded from summary, run not extended
`snapshot_omit()` exists (`riperf3/src/stream.rs:218`) but its **only** caller is a unit test (`stream.rs:1345`, inside `mod tests`). Production `client.rs`/`server.rs` never call it, and `omitted_errors`/`omitted_packets` are hardcoded to `0` in results (`client.rs:664`, `server.rs:524`).

**Current behavior:** `-O N` only tags the first N interval lines `(omitted)` in the reporter. The final summary divides cumulative bytes (including the warm-up) by the full `-t` duration (`client.rs` `build_results`, `:628-685`). The run also lasts only `-t` seconds total — omit is never added.

**iperf3 behavior:** runs for `omit + time`, resets statistics at the omit boundary, and reports throughput over `[omit, omit+time]` so slow-start is genuinely excluded.

**Impact:** the headline number diverges from iperf3 for any `-O` run — different window and shorter wall-clock. The "`-O` PASS" cell in BENCHMARKS only means the handshake completes, not that the statistic matches.

**Fix:** call `snapshot_omit()` at the omit boundary, subtract the omitted_* baselines in summary + results JSON, and extend the run by the omit period.

### LINKED ISSUE #171: Exchange retransmit totals include the omitted (-O) period once #162 lands
## Gap

PR #166 (#156) snapshots the sender's end-of-test retransmit total — connection-lifetime cumulative — into `StreamCounters` and ships it in the results exchange. iperf3 re-baselines `stream_retrans` at the omit boundary (`iperf_reset_stats` records `stream_prev_total_retrans` when `sender_has_retransmits == 1`, iperf_api.c:3687-3692), so its exchanged per-stream totals cover the post-omit window only.

Once #162 (real `-O`) merges, a riperf3 sender under `-O` with real retransmits will exchange a total that wrongly includes warm-up retransmits.

## Fix sketch

At the reporter's omit-boundary block (which already re-baselines reporter-side retransmit deltas), record a per-stream omit retransmit baseline next to the byte baselines in `StreamCounters`; the exchange build subtracts it from `final_retransmits`. Natural home: the #162 branch, which owns the boundary block.

## Notes

- Raised independently by PR #166 r2/r3 reviews and flagged on #162's thread; tracked here so whichever branch merges second picks it up.
- The reporter's `extremes.total_retransmits` has the same property (pre-existing).

therealevanhenry:
## Summary

`-O N` was cosmetic: it only tagged the first N interval lines while the run still lasted `-t` seconds and the summary divided cumulative bytes (warm-up included) by the full window. iperf3 runs `omit + time`, resets statistics at the boundary, and reports over the post-omit window only.

- **Run extension**: the client's duration arm sleeps `omit + time`; UDP senders' self-enforced `max_duration` covers it on both roles.
- **Reporter warm-up phase**: warm-up ticks emit `omitted` intervals on a `0..omit` timeline; the boundary is its own sleep arm (`-i 0`'s year-long ticker can't miss it), ordered before the ticker so a tick-coincident boundary (`-O 1 -i 1`) resolves as the final warm-up interval. At the boundary: warm-up tail flushed, every counter snapshotted (byte baselines, UDP `omitted_*`, retransmit baselines, end-block extremes), interval timeline restarted at 0, ticker re-phased.
- **Accounting**: summaries/`-J`/exchange use net (post-omit) byte getters; packets/errors stay **gross on the wire** with `omitted_*` baselines alongside, both sides subtract — exactly iperf3's exchange shape, which also makes riperf3 read a real iperf3 server's omit results correctly. `-n`/`-k` progress measures net, so the limit applies post-omit.
- **Validation**: `build()` rejects `omit > 60` with iperf3's IEOMIT wording.

## TDD

Red commit first: `-O 1 -t 2` measured 2.26 s wall (vs required ~3 s) and 26.2 GB reported vs 12.9 GB in post-omit intervals — both tests now green, plus `test_start.omit`, `omitted:true` flags, timeline restart at 0, and summary internal consistency. Full workspace suite green (438), incl. the adjacent #55 final-flush and #103 measured-window suites.

## Gate before merge

`/riperf3-verify` vs real iperf3 `-O` in both roles on the fleet (results to be posted here), plus the standard post-merge smoke.

Closes #31.

Fixes #31. Fixes #171 (exchanged retransmit totals re-baselined at the omit boundary — landed here per the #166 review agreement).

### DIFF
--- a/riperf3-cli/tests/omit_behavior.rs
+++ b/riperf3-cli/tests/omit_behavior.rs
@@ -99,7 +99,7 @@ fn omit_extends_run_and_resets_timeline() {
             "1",
             "-J",
         ],
-        Duration::from_secs(25),
+        Duration::from_secs(60),
         "client",
     );
     let _ = server.0.wait();
@@ -169,7 +169,7 @@ fn omit_summary_excludes_warmup_bytes() {
             "1",
             "-J",
         ],
-        Duration::from_secs(25),
+        Duration::from_secs(60),
         "client",
     );
     let _ = server.0.wait();
@@ -211,7 +211,7 @@ fn omit_with_byte_limit_applies_post_omit() {
 
     let (out, elapsed) = run_capturing(
         &["-c", "127.0.0.1", "-p", &ps, "-O", "1", "-n", "100M", "-J"],
-        Duration::from_secs(25),
+        Duration::from_secs(60),
         "client",
     );
     let _ = server.0.wait();
@@ -222,10 +222,54 @@ fn omit_with_byte_limit_applies_post_omit() {
     );
     let v: Value =
         serde_json::from_str(&out).unwrap_or_else(|e| panic!("client -J invalid ({e}): {out}"));
+    // `-n 100M` is binary (unit_atoi): 104,857,600. The budget design lands
+    // exactly on it; allow one 128 KiB block of slack. The old driver-poll
+    // refill overshot 1.2-4x (review r2 blocker 3).
     let bytes = v["end"]["sum_sent"]["bytes"].as_u64().expect("bytes");
     assert!(
-        bytes >= 95_000_000,
-        "the -n budget must be transferred POST-omit, got {bytes}: {out}"
+        (104_857_600..=104_988_672).contains(&bytes),
+        "post-omit -n must land on the 104857600-byte target (r2): {bytes}: {out}"
+    );
+}
+
+/// Reverse `-R -n + -O` (r2 blocker 1): the SERVER's senders hit the same
+/// budget design — pre-fix they exhausted the gross budget during warm-up and
+/// broke permanently, so the client's post-omit net target was unreachable
+/// (infinite zero-byte intervals).
+#[test]
+fn omit_with_reverse_byte_limit_terminates() {
+    let port = free_port();
+    let ps = port.to_string();
+    let mut server = spawn_server(&ps);
+
+    let (out, elapsed) = run_capturing(
+        &[
+            "-c",
+            "127.0.0.1",
+            "-p",
+            &ps,
+            "-R",
+            "-O",
+            "1",
+            "-n",
+            "50M",
+            "-J",
+        ],
+        Duration::from_secs(60),
+        "client",
+    );
+    let _ = server.0.wait();
+
+    assert!(
+        elapsed < Duration::from_secs(20),
+        "reverse -n + -O must terminate (r2 blocker 1)"
+    );
+    let v: Value =
+        serde_json::from_str(&out).unwrap_or_else(|e| panic!("client -J invalid ({e}): {out}"));
+    let bytes = v["end"]["sum_received"]["bytes"].as_u64().expect("bytes");
+    assert!(
+        (47_500_000..=53_000_000).contains(&bytes),
+        "reverse post-omit budget must land on ~50M, got {bytes}: {out}"
     );
 }
 
@@ -260,7 +304,7 @@ fn server_intervals_rebased_after_omit() {
             "-i",
             "1",
         ],
-        Duration::from_secs(25),
+        Duration::from_secs(60),
         "client",
     );
     let mut sout = String::new();
--- a/riperf3-cli/tests/omit_behavior.rs
+++ b/riperf3-cli/tests/omit_behavior.rs
@@ -1,0 +1,200 @@
+//! CLI integration tests: `-O/--omit` must behave like iperf3 (#31), not just
+//! tag interval lines. iperf3 runs for `omit + time` seconds, resets statistics
+//! at the omit boundary (the interval timeline restarts at 0), and reports the
+//! summary over the post-omit window only — so slow-start is genuinely
+//! excluded. Pre-#31 riperf3 ran only `-t` seconds and divided cumulative
+//! bytes (including warm-up) by the full duration.
+
+use std::io::Read;
+use std::process::{Command, Stdio};
+use std::time::{Duration, Instant};
+
+use serde_json::Value;
+
+fn free_port() -> u16 {
+    std::net::TcpListener::bind("127.0.0.1:0")
+        .expect("bind ephemeral port")
+        .local_addr()
+        .expect("local_addr")
+        .port()
+}
+
+struct ChildGuard(std::process::Child);
+impl Drop for ChildGuard {
+    fn drop(&mut self) {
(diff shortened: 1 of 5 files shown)


therealevanhenry:
## Fleet verify vs real iperf3 (gate): PASS

| cell | result |
|---|---|
| riperf3↔riperf3 `-O 2 -t 4` | wall **6.21 s** (omit+time ✓), window 0–4 s, 63.8 Gbit/s |
| riperf3 → **iperf3 server** | wall 6.1 s; iperf3's receiver total (34829 MB) matches our post-omit sender (34821 MB) — exchange `omitted_*` interop ✓ |
| **iperf3 client** → riperf3 server | 33.1 GB both sides agree, 71 Gbit/s — riperf3's server-side omit accounting reads correctly ✓ |
| riperf3 UDP `-u -b 100M -O 1 -t 2` | wall 3.21 s (omit+time ✓), window 2 s ✓ |

(iperf3 reference for the same TCP params: wall 6.01 s, window 0–4 s.) A pre-existing UDP `-b`-at-moderate-rates anomaly was observed in the control run **with and without** `-O` — filed separately, not a #31 regression.

therealevanhenry:
r2 blockers addressed via the suggested redesign: senders pause-at-limit (iperf3 mt semantics, also fixing the warm-up profile divergence in finding 4), refill in the reporter's boundary block (baseline+refill coincide — post-omit -n is now byte-exact, upper bound pinned), server-side budget wired through its reporter (reverse hang gone, pinned by a new test), third subtraction site fixed. Ready for r3.

therealevanhenry:
Round-3 findings addressed (d750b2c + merge c1e2f94 + cde1080). Answering the semantics question first, since it drove the design:

**[QUESTION] → gross-received semantics, on record.** Per the faithful-drop-in ethos: the receive-side `-n`/`-k` limit counts **gross** bytes. `iperf_reset_stats` zeroes `test->bytes_sent` (iperf_api.c:3675) but never `test->bytes_received`; the end check (iperf_client_api.c:771-772) is `sent >= n || received >= n`. So a reverse/bidir `-O` run whose warm-up already covered the limit ends AT the boundary — verified against iperf3 3.20 directly: `--bidir -n 50M -O 1` → `sent: 1310720 recv: 0 secs: 0.0`; `-R -n 20M -O 1` → `recv: 0 secs: 0.0`. riperf3 now produces the same shape. The old reverse test pinned the unfaithful fresh-N semantics and has been rewritten (`omit_with_reverse_byte_limit_ends_on_gross_bytes`, paced with `-b 20M` for CI stability so the post-omit net lands deterministically at ≈ target − warm-up); a paced bidir termination test was added alongside.

**[BLOCKER 1] fixed structurally.** `transferred_bytes` is now asymmetric: senders sum `bytes_sent_net()`, receivers sum **`bytes_received()` (gross)**. Gross received is monotonic, so the end check cannot race either reporter's boundary baselines — the hang class is dead by construction, not by tightened timing. Empirical: reverse self-to-self 12/12 clean (was 7/12 hang), vs real iperf3 servers 6/6 + 3/3 clean (was 6/6 hang), forward `-n 100M [-P 8] -O 1` still byte-exact (104,857,600).

**[BLOCKER 2] fixed** — the two test call sites are fmt-clean (caught locally as the exact diff the review predicted).

**[BLOCKER 3 / races A+C] fixed.** Boundary block order is now baselines → budget refill → signal. The `-n`/`-k` driver no longer gates on a parallel wall clock: `wait_byte_limit` awaits a new `OmitBoundary` Notify the reporter fires at the END of its boundary block, then runs its first end check before any sleep (iperf3's select loop re-checks per wake; a 100 ms first poll would leak a poll of line rate into the post-omit window). A wall fallback (omit + 2 s) preserves liveness if the reporter dies pre-boundary.

**[NIT 1] (#147 interaction):** J (#166) merged first and this branch is updated onto it. The pause loop re-enters `while !done` every 10 ms, and `run()`'s DoneOnDrop sets `done` on every exit path including `?` propagation — so error paths can't park a sender forever in lib-consumer runtimes. One real consequence surfaced: J's `exchange_retransmits_survive_sender_socket_close` awaited a sender's self-termination at budget exhaustion, which pause-at-limit semantics turned into an infinite idle — the merged test now drives `done` exactly like the production driver (caught by an 8-hour suite wedge; the suite now runs under a hard timeout).

**[NIT 2] fixed** — the claim is a no-undo `fetch_update` (claims only from positive budget), eliminating the undo-lands-on-refilled-budget leak; total overshoot is bounded below one block.

**[NIT 3] fixed** — `build()` rejects `-i` outside {0} ∪ [0.1, 60] with the IEINTERVAL message, red-first tested.

**Ride-along: #171** (filed from the #166 review, agreed to land in whichever branch merged second — this one): the boundary block stores each sender's cumulative retransmit count into `StreamCounters` (iperf3's `stream_prev_total_retrans`); both roles' exchange builds subtract it, so exchanged totals cover the post-omit window. Red-first unit test.

Gates: fmt/clippy clean, 465/465 workspace tests (hard 480 s timeout), 7-target xcheck PASS, release-build repro battery above. Ready for round 4.

therealevanhenry:
Round-4 verdict was APPROVE; the three hardening nits are applied in 9c40709 rather than deferred, since they touch the same concurrency-critical lines this PR owns:

- **Nit 1:** `OmitBoundary` fast path upgraded to `Release`/`Acquire` so a driver that skips the Notify still observes the boundary block's baseline/refill stores on weakly-ordered targets.
- **Nit 2:** the server's refill target is now captured at budget creation, before any sender can consume in the TestStart→reporter-spawn gap — the invariant is explicit instead of accidentally-never-binding.
- **Nit 3:** the claim-site comment now distinguishes the <1-block *budget* overshoot from the additional one in-flight block per sending stream in recorded net (`-P k` can land at N + k blocks), so nobody pins the 1-block figure in a future test.

Nit 4 (no control-socket watch during `-n`/`-k`) is pre-existing on main and already tracked as the third item of #170.

Post-APPROVE delta is two comments, one memory-ordering upgrade, and one load hoist; gates re-run green on 9c40709 (fmt/clippy, 465/465 with hard timeout, reverse/forward repro sanity 4/4 + byte-exact). Merging on the r4 APPROVE.

Issue #122, opened 2026-06-08
Builder exposes inert setters the library never consumes (affinity, dscp, get_server_output, pidfile, logfile)


therealevanhenry:
Surfaced by #43: once `Client`/`Server` fields became `pub(crate)`, dead-code analysis (previously suppressed by `pub`) flagged 7 fields as **never read**:

- **Client**: `get_server_output`, `affinity`, `dscp`, `pidfile`, `logfile`
- **Server**: `pidfile`, `logfile`

Each has a builder setter (`ClientBuilder::affinity()`, etc.) the **CLI calls**, but the library never acts on the stored value — the CLI realizes these at the process level from its own `Cli` struct:
- `affinity` → CLI calls `riperf3::set_cpu_affinity()` directly (`main.rs:82`), not via the field.
- `pidfile`/`logfile` → CLI writes the pidfile / dup2's stdout itself, before building the runtime (`main.rs:54-72`). A *library* arguably shouldn't daemonize at all.
- `dscp` → never applied by `Client::run()` (only `tos` is). A library consumer calling `.dscp()` gets no effect.
- `get_server_output` → never read; the client's server-output fetch is unconditional. The flag is a **no-op** — possible latent bug (distinct from the others).

For #43 these are annotated `#[allow(dead_code)]` to keep that PR a clean visibility change.

## Decision needed (per field)
Two reasonable resolutions, and it differs by field:
1. **Prune** the field + builder setter + the redundant CLI setter call. Correct for `pidfile`/`logfile` (process-level, not library concerns) and arguably `affinity` (CLI already does it).
2. **Wire into the library** — make `Client::run()` actually apply it. Correct if a pure-library consumer should get the behavior: `dscp` (apply alongside `tos`), `get_server_output` (gate the server-results fetch).

Breaking either way (removes builder methods or changes behavior); fine to land in 0.7.0 while unpublished, or defer to a later major. Filed rather than bundled to keep #43 scoped to encapsulation.

therealevanhenry:
**Resolution (0.7.0):** split into prune vs wire.

- **Pruned** in #128 (breaking, 0.7.0): `ClientBuilder::{pidfile, logfile, affinity}` and `ServerBuilder::{pidfile, logfile}` removed — the CLI realizes these at the process level, so they were inert library surface.
[...]

Issue #47, opened 2026-05-29
Tracking: skeptical post-0.5.4 review follow-ups (#31–#46)


therealevanhenry:
Follow-ups from a skeptical post-0.5.4 review of the project's headline claims (idiomatic Rust, wire-compatible iperf3 replacement, cross-platform parity, performance). Grouped by theme; each item is a standalone issue.

### Interop / behavioral fidelity (the "replacement" gap)
- [x] #31 `-O`/omit is cosmetic — warm-up not excluded from summary, run not extended
- [x] #32 `--pacing-timer` parsed but never wired (silent no-op)
- [x] #33 `--get-server-output` silent no-op (never sent, never printed)
- [x] #34 `-T`/`--title` doesn't prefix output lines
- [x] #35 `--extra-data` never emitted in JSON
- [x] #36 `-J` output not faithful to iperf3 schema (empty intervals, fabricated host:port); rebuild from typed structs
- [x] #37 `congestion_used` never populated — applied CC not reported

### Verification gap (claims not gated by CI)
- [x] #38 Add a real-iperf3 interop gate to CI (suite is loopback-only — a wrong wire constant would still pass all tests)
- [ ] #39 CI only `cargo check`s macOS/Windows; FreeBSD has no CI

### Cross-platform correctness / docs
- [ ] #40 macOS advertises retransmit info but always reports 0
- [ ] #41 Right-size cross-platform claims in README ("full parity" overstated)

### Code quality / correctness
- [x] #42 `-Z` zerocopy temp file uses a fixed per-process name; races under `-P>1`
- [ ] #43 Collapse builder boilerplate; `Client` fields shouldn't all be `pub`
- [ ] #44 Unsafe audit table omits `getsockopt(TCP_MAXSEG)`
- [ ] #45 Inconsistent error handling for socket options (`?` vs `let _ =`)
- [ ] #46 README test count stale (says 269; actual 334)

**Note on scope:** two review recommendations were folded into their parent defect rather than filed separately — "model `-J` with typed serde structs" lives in #36, and "make a real iperf3 a test dependency" lives in #38. Split them back out if you'd rather track the refactor independently.

therealevanhenry:
**Policy (maintainer):** riperf3's ethos is *faithful implementation* — a fully wire-compatible, drop-in replacement for iperf3, in idiomatic Rust.

[...]

PR #164, opened 2026-06-09
feat(protocol): wire --get-server-output end to end (#33)

### LINKED ISSUE #33: --get-server-output is a silent no-op (never sent to server, never printed)
`--get-server-output` is plumbed to `Client.get_server_output` and the CLI builder (`riperf3-cli/src/main.rs:163`), but `build_params` never sets `p.get_server_output` (`riperf3/src/client.rs:233-296`), so the server is never told to return its output, and the client never fetches/prints any server-side text.

**iperf3 behavior:** returns the server's console output to the client at the end of the test.

**Fix:** send the param, have the server include its output in the results exchange, and print it client-side — or reject the flag if it won't be implemented.

therealevanhenry:
## Summary

`--get-server-output` was a silent no-op (#33): the param was never sent, nothing fetched or printed. Now faithful to iperf3:

- Client sends `get_server_output: 1`; the server attaches **`server_output_text`** (text mode — console report diverted into a capture sink, mirroring iperf3's tmpfile diversion; the server's own stdout stays silent for the test) or **`server_output_json`** (a `-J` server attaches its full report, built once pre-exchange and reused at print — `build_report_input` drains the interval collector destructively, so build is split from print).
- Client prints `Server output:` + text after its own report, or carries the keys at the `-J` report tail (iperf3's placement after `end`). Exchange fields are additive optional serde fields (`skip_serializing_if` + `default`) — wire- and semver-safe with peers that lack them, including iperf3 servers without the flag.
- Capture sink lives beside `OUTPUT_TITLE` in macros.rs (RAII, clears on drop, shares the documented process-global caveat) and diverts both `vprintln!` and the reporter's `titled` printers.

**Scoped out:** `--json-stream` × get-server-output (iperf3's streamed shape needs live verification first) — will file on merge.

## TDD

Red commit first: text/text (client prints the section, server console diverted), `-J` client (`server_output_text`), `-J`/`-J` (`server_output_json` with start/intervals/end) — all green now; the no-flag control pinned absence throughout. 442 workspace tests green, clippy/fmt clean.

## Gate

`/riperf3-verify` vs real iperf3 both roles before merge (the server-text composition is the live-verify item), plus the post-merge smoke. Closes #33; #122 and #47 close after merge.

### DIFF
--- a/riperf3-cli/tests/get_server_output.rs
+++ b/riperf3-cli/tests/get_server_output.rs
@@ -1,8 +1,8 @@
 //! CLI integration tests: `--get-server-output` must work like iperf3 (#33) —
 //! the server returns its console output (text mode) or its full `-J` report
-//! (JSON mode) in the results exchange, and the client prints/attaches it.
-//! Pre-#33 the flag was a silent no-op: the param was never sent and nothing
-//! was fetched or printed.
+//! (JSON mode) in the results exchange, and the client prints/attaches it —
+//! while the server console stays live (iperf3 dual-writes; it never
+//! diverted). Pre-#33 the flag was a silent no-op.
 #![cfg(unix)]
 
 use std::io::Read;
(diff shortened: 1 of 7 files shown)


therealevanhenry:
Round-1 findings addressed: **B1** — capture is now a TEE (console stays live; iperf3 has dual-written since 2014, the tmpfile premise was wrong; test flipped to pin dual-write). **B2** — 'Server JSON output:' + trailing blank line. **N3** — filed as #168, referenced at the scope-out site. **N4** — doc blocks un-mangled. **N6** — client print/attach gated on having requested the flag. **N5** (timestamps-in-capture) folded into #168. **N7** — no action (noted as more-self-consistent divergence). 442 green.

Issue #137, opened 2026-06-09
Enhancement: Return Rich JSON Report from `Client::run()` for Library Users


Sean-Bartie:
## Description
Enhance the library return value of `Client::run()` so library consumers can receive the full JSON report object (the same rich structure used for `-J` output), rather than only the server-side `TestResultsJson` exchanged during control-channel result exchange.

## Problem
Today, library users who call `Client::run()` and inspect the result are limited by `TestResultsJson`.

`TestResultsJson` is better suited as an internal control-channel exchange model than as the primary library-facing result type.

`TestResultsJson` is useful, but it does not include the full set of metrics and structured fields available in the JSON report model (start metadata, interval arrays, richer end summaries, per-direction aggregates, and related report details).

## Why this matters
- Better parity between CLI output and library API.
- Improved observability and downstream integrations (dashboards, telemetry pipelines, custom analytics).
- Simpler application code for users embedding riperf3.

Sean-Bartie:
Enhancement of https://github.com/therealevanhenry/riperf3/pull/2

therealevanhenry:
Thanks @Sean-Bartie — agreed, and accepted for the next **major** release (0.8.0). `TestResultsJson` was the right v1 (your PR #2), but as a control-channel exchange model it shouldn't stay the primary library-facing result. Design we've settled on:

## 0.8.0 API

```rust
// 0.7.x
let results: TestResultsJson = client.run().await?;

// 0.8.0
let results = client.run().await?;
results.report           // json_report::Report — the full -J-schema report, now ALWAYS built
results.server_exchange  // the raw TestResultsJson the server sent (the 0.7.x value)
```

`Client::run()` returns a `#[non_exhaustive] TestResults { report, server_exchange }` wrapper rather than a bare `Report`, for two reasons:

1. **The exchange isn't fully subsumed by the report.** It carries server-side data the `-J` report deliberately doesn't surface (`sender_has_retransmits`, the *server's* `congestion_used`, raw server per-stream stats in bidir) — returning only `Report` would silently drop the data your PR #2 exposed.
2. **Future-proofing**: the wrapper gives later releases a place to add fields without another breaking bump.

[...]

Issue #102, opened 2026-06-05
TCP client ignores -b/--bitrate (runs unthrottled); iperf3 caps it


therealevanhenry:
Surfaced during #101 review (pre-existing, not introduced there).

`riperf3 -c <host> -b 1G` (TCP) runs at full line rate (~120 Gbit/s on loopback) instead of pacing to 1 Gbit/s. iperf3 caps a TCP client at the requested `-b`. UDP `-b` IS honored in riperf3, so this is TCP-path-specific.

Faithful behavior: pace the TCP send loop to the target bitrate (iperf3 uses a token-bucket / timer between writes). Adjudicate against iperf3's `iperf_check_throttle`/`iperf_send` pacing. Note riperf3's existing UDP pacing as the reference for the rate model.

Issue #18, opened 2026-05-29
UDP sender polish: fatal-error busy-loop, dead clamp, --sendmmsg on macOS


therealevanhenry:
Minor, pre-existing UDP sender polish surfaced during the #6 cold reviews. None
are correctness bugs; bounded or cosmetic. Grouped for triage.

1. **Non-fatal send errors busy-loop until the deadline.** In both UDP senders
   (`stream.rs`, the per-packet `Err(e)` arm and the sendmmsg non-EAGAIN `Err(e)`
   arm), a persistent immediate error — e.g. `ECONNREFUSED` surfaced on a
   connected UDP socket after an ICMP port-unreachable — does `seq -= …` and
   re-loops with no sleep, spinning a core until `-t`. Bounded by the deadline
   and matches iperf3's general posture, but a small backoff or break-on-fatal
   would be cleaner.

2. **Dead clamp.** `stream.rs` per-packet sender: `batch_size = 32_u64.clamp(1, 64)`
   is a constant `32` — the clamp is a no-op on a literal. Simplify to `32`.

3. **`--sendmmsg` on macOS silently degrades.** The flag is rejected only under
   `#[cfg(not(unix))]`, but the real `sendmmsg` impl is gated to
   linux/freebsd/netbsd. On macOS the flag is accepted and falls back to the
   per-packet blocking sender — graceful, but the "error at parse time, matches
   iperf3" framing isn't accurate there. Either reject on macOS too or adjust
   the wording.

Severity: low. Filed for tracking, not blocking.

PR #165, opened 2026-06-10
docs: 0.7.2 governance + accuracy ride-alongs

therealevanhenry:
## Summary

The non-code ride-alongs for 0.7.2 (from the holistic project review):

- **README accuracy**: "369 tests" → 430+ (tree has 439 test fns and growing); platform wording right-sized to what CI actually enforces — the old text claimed FreeBSD had no CI, stale since the required FreeBSD job landed; macOS/Windows are full native suites, informational.
- **CONTRIBUTING.md**: fidelity ethos (cite iperf3 source for behavioral claims), tests-before-fixes, the full local gate (fmt/clippy/test + the 7-target cross-check loop), CI gate inventory, semver/release discipline, lib-API-only convention.
- **SECURITY.md**: GitHub private vulnerability reporting + email fallback; scope notes (untrusted control-channel parsing, RSA auth path, unsafe inventory pointer).
- **MSRV declared**: `rust-version = "1.85"` — the floor is set by the dependency tree (locked base64ct 1.8 via the rsa chain is edition 2024, needing Cargo >= 1.85); our own code's floor is 1.82 (`Option::is_none_or`). Inherited by both crates, proven empirically (1.85 passes, 1.84 fails on base64ct) and enforced by a new pinned-toolchain CI `cargo check` job.
- **rustdoc for all 71 builder setters** (56 client, 15 server): each names its iperf3 flag and load-bearing semantics, grounded in the CLI wiring. `cargo doc` zero warnings — docs.rs UX for #137's library audience before 0.8.0 stabilizes this surface.

Docs/metadata only — no behavior change (the MSRV CI job is additive).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,17 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo test --workspace
 
+  msrv:
+    name: MSRV (1.77)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@1.77
+      - uses: Swatinem/rust-cache@v2
+      # Check (not test) on the declared floor: proves the rust-version claim
+      # in Cargo.toml without doubling the test wall-clock.
+      - run: cargo check --workspace --all-targets
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,11 +20,11 @@ jobs:
       - run: cargo test --workspace
 
   msrv:
-    name: MSRV (1.77)
+    name: MSRV (1.85)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.77
+      - uses: dtolnay/rust-toolchain@1.85
       - uses: Swatinem/rust-cache@v2
       # Check (not test) on the declared floor: proves the rust-version claim
       # in Cargo.toml without doubling the test wall-clock.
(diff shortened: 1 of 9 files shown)


therealevanhenry:
r1 fixes pushed (in 59f7e86 — commit subject is mislabeled from a tooling mixup; the squash subject will be correct): MSRV → **1.85** (lockfile v4 + base64ct edition-2024; own-code floor 1.82 via is_none_or) with the CI job repinned; Windows-required wording fixed everywhere + the stale ci.yml comments corrected; congestion/gsro doc phrasing now Linux/FreeBSD-honest; rcv_timeout doc notes SO_RCVTIMEO doesn't fire on nonblocking sockets; **private vulnerability reporting enabled** on the repo so SECURITY.md's primary channel works. On the inert-option docs (r1 blocker 4): #162 (omit) and #164 (get-server-output) implement exactly those behaviors and both merge BEFORE this PR — at merge time the docs are accurate; sequencing is the gate.

therealevanhenry:
Round-2 findings addressed in 736a8b5:

- **Platform lists (blocker):** the r1 fix had pasted one "Linux/FreeBSD only" string onto four setters; each now states its verified gate. `congestion` was already correct. `zerocopy` → Linux/macOS/FreeBSD, with the documented fallback to the copying sender elsewhere and under `-b`/`-n`/`-k` (matches the #102-era dispatch in client.rs). `gsro` → Linux only, silent no-op elsewhere (net.rs cfg gates). `bind_dev` → `SO_BINDTODEVICE` on Linux, `IP_BOUND_IF`/`IPV6_BOUND_IF` on macOS, silent no-op elsewhere.
- **Cargo.toml MSRV comment:** corrected — the 1.85 floor stands on base64ct (edition 2024) alone; the v4 lockfile parses on Cargo >= 1.78 and was wrongly cited.
- **Nits:** `snd_timeout` now carries the "(Linux only)" tag like its siblings; PR body's stale `1.77`/`offset_of!` rationale replaced with the verified 1.85/base64ct one.

Sequencing unchanged: this PR merges after #162 and #164 so the omit/get-server-output docs land true.

therealevanhenry:
Round-3 finding addressed in 3a9bf5c:

- **Restored the true clause (blocker):** the r2 rewording deleted "rejected at `build()` on non-unix" along with the false platform list — but `build()`'s `#[cfg(not(unix))]` block (client.rs:1777-1799) really does reject `zerocopy`/`bind_dev`/`congestion`/`gsro` with `ConfigError::Unsupported`. All three reworded setters now state both halves: silent on other unix, rejected at `build()` on non-unix (zerocopy additionally keeps its `-b`/`-n`/`-k` fallback note).
- **Nit:** README's "Unsupported flags return a clear error at startup" tightened to distinguish the Windows `build()` error from the silent Unix no-op for blank table cells.

Round 4 should be the trivial re-check the r3 reviewer anticipated.

therealevanhenry:
Round-4 finding addressed: the silent-no-op claim is now scoped to the socket layer, with `-D` and `--sendmmsg` carved out as erroring at startup wherever unsupported (daemon check in the CLI, sendmmsg's deliberate #18 rejection at `build()`).

Issue #156, opened 2026-06-09
Results exchange never sends sender_has_retransmits=1 — peer's Retr display suppressed cross-tool


therealevanhenry:
Surfaced during the #152 round-1 review (pre-existing; #152 makes it more pointed now that macOS also has retransmit info).

riperf3 hardcodes the results-exchange `sender_has_retransmits` to `0` when sending and `-1` otherwise (client.rs ~851, server.rs ~626) and never consults `tcp_info::has_retransmit_info()`. iperf3 sends **1** when it is a sender on a retransmit-capable platform (`check_sender_has_retransmits`, iperf_api.c:634-640), and the **peer** uses the received value to gate display of the remote sender's Retr column (iperf_api.c:2855-2857, 4261).

Net effect: a riperf3 sender talking to an iperf3 receiver suppresses the iperf3 side's Retr rendering on every platform, including Linux — a wire-visible fidelity gap.

**Fix:** set `sender_has_retransmits = if is_sender && tcp_info::has_retransmit_info() { 1 } else { 0/-1 per iperf3's convention }` on both roles; verify the exact -1/0/1 semantics against iperf3's reader before wiring; add an interop assertion to the compat matrix run (riperf3 client → iperf3 server `-V` shows Retr).

0.7.2-eligible: small wire-value fix, gated by the interop matrix.

therealevanhenry:
Fixed in 0.7.2 by PR #166: sender_has_retransmits is set on both roles (client.rs/server.rs), with a test pin.

Issue #155, opened 2026-06-09
FreeBSD Cwnd inflated ~mss×: tcpi_snd_cwnd is bytes, reader multiplies by mss


therealevanhenry:
Surfaced during the #152 round-1 review (pre-existing; same unit-bug class #152 fixed on macOS).

FreeBSD's `tcp_usrreq.c` fills `tcpi_snd_cwnd = tp->snd_cwnd` — **bytes** — and iperf3's `get_snd_cwnd()` FreeBSD branch returns it raw. riperf3's FreeBSD reader (`tcp_info.rs`, `snd_cwnd: info.tcpi_snd_cwnd as u64 * info.tcpi_snd_mss as u64`) therefore inflates the Cwnd column (text + `-J` `max_snd_cwnd`) by roughly mss× vs iperf3 on FreeBSD.

**Fix:** drop the multiplication in the FreeBSD reader (one line) + verify a plausible-range Cwnd in the FreeBSD CI test run (the #152 macOS test pattern, `macos_snapshot_reads_sane_values`, ports directly).

0.7.2-eligible: tiny, faithful, FreeBSD CI is a required check.

therealevanhenry:
Fixed in 0.7.2 by PR #166: FreeBSD snd_cwnd is converted to bytes in tcp_info.rs.

Issue #142, opened 2026-06-09
UDP interval-sum jitter should be the average across receiving streams (-P ≥ 2), not the last stream's


therealevanhenry:
Surfaced during the #141 cold review (pre-existing; the old lumped sum had the same behavior, and #141's per-direction split preserved it 1:1).

iperf3 reports a receiving direction's per-interval sum jitter as the **average across that direction's streams**: `avg_jitter /= test->num_streams` (`iperf_api.c` `iperf_print_intermediate`, ~:4042 @ 8701e52). riperf3's `DirAcc::last_jitter` (reporter.rs) carries the **last receiving stream's** jitter instead — identical at `-P 1`, divergent at `-P ≥ 2` UDP.

**Fix:** accumulate `jitter_sum` + receiving-stream count in `DirAcc` and emit the mean in `direction_interval_sum`. Check whether the text `[SUM]` line and the final summary's aggregate jitter (`sum_summaries` uses max — verify against iperf3's end-block jitter semantics separately before touching it) share the defect.

therealevanhenry:
Fixed in 0.7.2 by PR #166 (verified during the 0.7.3 campaign recon): interval-sum jitter is the per-direction mean in reporter.rs's sum_summaries, with a test pin.

PR #166, opened 2026-06-10
fix: review-findings batch (#142, #147, #155, #156)

therealevanhenry:
Four small faithfulness/correctness fixes discovered by this release's cold reviews:

- **#142** — interval-sum jitter is the **mean** across receiving streams (iperf3 `avg_jitter`), not last-wins; diverged at `-P ≥ 2` UDP.
- **#147** — `ServerTerminate` mid-test leaked the reporter task + running senders into a library consumer's runtime (the early return skipped `finish`/`done`). The abort path now shuts down before propagating. Pinned by a full-handshake mock terminating mid-test, asserting post-abort flow is buffered-only (single-digit MB) vs the leak's line-rate GBs.
- **#155** — FreeBSD `tcpi_snd_cwnd` is bytes; the `×mss` inflated Cwnd ~1500×. Same class as #152's macOS fix; FreeBSD required CI exercises the reader.
- **#156** — exchange now sends `sender_has_retransmits=1` for retransmit-capable TCP senders (both roles) per `check_sender_has_retransmits`; the peer gates *our* Retr display on it, so the old `0` suppressed riperf3 senders' Retr in iperf3 receivers on every platform. Interop-matrix verified at the release gate.

445 workspace tests green, clippy/fmt clean.

### DIFF
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -858,13 +858,20 @@ impl Client {
                 // #156: when sender_has_retransmits=1, the peer prints this
                 // value (iperf3 get_results stores it into stream_retrans and
                 // the summary renders it) — it must be the REAL end-of-test
-                // total, read directly from the kernel, not a sentinel.
+                // total. The sender task snapshots it into the counters while
+                // its socket is still open; by exchange time the socket is
+                // closed, so a live fd read here only serves as a fallback
+                // for paths that never reached the snapshot.
                 let retransmits =
                     if s.is_sender && crate::tcp_info::has_retransmit_info() && !is_udp_stream {
-                        s.raw_fd
-                            .and_then(crate::tcp_info::get_tcp_info)
-                            .map(|i| i.total_retransmits as i64)
-                            .unwrap_or(-1)
+                        match s.counters.final_retransmits() {
+                            n if n >= 0 => n,
+                            _ => s
+                                .raw_fd
+                                .and_then(crate::tcp_info::get_tcp_info)
+                                .map(|i| i.total_retransmits as i64)
+                                .unwrap_or(-1),
+                        }
                     } else {
                         -1
                     };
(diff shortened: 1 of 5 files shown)


therealevanhenry:
Round-1 findings addressed in f1c102e:

- **#156 wire value (blocker):** setting `sender_has_retransmits: 1` obligated us to send real per-stream retransmit totals in the results exchange — iperf3 peers print our `streams[].retransmits` when the flag is 1, and we were hardcoding `-1` (serialized as `u64::MAX`). Both roles now read the kernel total via `get_tcp_info` per sender stream at exchange-build time (`build_results` client-side, the server mirror), on platforms where `has_retransmit_info()` holds.
- **#147 test (blocker):** the e2e mock pinned nothing — `DoneOnDrop` already guaranteed sender shutdown, so the mock passed with or without the fix. Replaced with `abort_path_joins_the_reporter`, which calls `run_test` directly (bypassing the drop guard) and discriminates on `Arc::strong_count(&collector) == 1` after the abort path returns — a parked reporter task holds a clone, so the test is red without the fix. The old mock is kept as a commented sanity sketch.

Follow-ups out of scope for this batch, filed separately:
- #169 — END-block summary jitter is also the mean in iperf3 (`avg_jitter += sp->jitter` at `iperf_api.c:4249`, `/= num_streams` at `:4490`); this PR fixed the interval level (#142), the END block has the same bug one layer up.
- #170 — three data-phase gaps vs iperf3's select loop (control-death accounting, ServerTerminate summary parity, no control watch under `-n`/`-k`).

therealevanhenry:
Round-2 blocker addressed in 15db8ce (branch also updated to main, which now includes #153/#157/#164):

- **#156 totals survive the socket close:** every TCP sender (`run_tcp_sender` + the three `run_tcp_sender_zerocopy` variants) now snapshots the kernel's retransmit total into `StreamCounters` immediately before returning — i.e. while the socket is provably open. `build_results` (both roles) reads the snapshot; the live fd read remains only as a fallback for paths that never reached the snapshot. This removes the EBADF race outright rather than narrowing it, and sidesteps the recycled-fd hazard the review flagged (a recycled fd can no longer report another socket's totals, because the snapshot wins).
- **Red-first test:** `exchange_retransmits_survive_sender_socket_close` runs a real loopback sender to completion (socket dropped), then asserts `build_results` still reports a real total beside `sender_has_retransmits: 1`. It models the dead-fd state with `raw_fd: None` deterministically — a real closed fd can be recycled by a parallel test's socket, making `get_tcp_info` spuriously succeed; the pinned property (no exchange-time fd dependence) is identical. Red on the pre-fix tree (`got -1`), green post-fix.
- **Nits:** stale FreeBSD wording on `TcpInfoSnapshot::snd_cwnd` corrected; `DirAcc::udp_recv` dropped in favor of `udp_recv_count > 0`.
- Gates: fmt/clippy clean, 455/455 workspace tests, 7-target cross-check all PASS.

Note for the G (#162) integration: iperf3 re-baselines `stream_retrans` at the omit boundary (`iperf_reset_stats`), so when #162 lands, the snapshot here must subtract a boundary baseline. Flagged in #162's review thread; whichever PR merges second picks it up at update-branch time.

PR #176, opened 2026-06-10
test: retry clients on ConnectionRefused; sub-ephemeral free_port (fixes the UDP bidir CI flake)

therealevanhenry:
Fixes the recurring `udp_bidir_interval_sums_split_udp_stats` CI flake (PR #166's CI and main run 27281845954, both with empty client stdout).

**Root cause:** the harness waited a fixed 2 s for spawned servers to bind; on loaded runners the (debug, cold) server binds later, the client dies on ECONNREFUSED, and with stderr nulled and exit status ignored the only symptom was `not valid JSON (EOF at line 1 column 0)`.

**Fix (third design — the first two died in review/CI and are documented in `tests/common/mod.rs` so nobody re-walks them):**
- **Client retry on `ConnectionRefused`** (`common::run_client`/`run_client_ok`): retries while the run exits unsuccessfully with `ConnectionRefused` in stderr (10 s window, 100 ms backoff, hard per-attempt timeout, per-attempt `elapsed` so timing assertions stay honest). Race-free by construction: refused connects never reach `accept()` (one-off servers untouched), nothing ever holds the server's port, and the kind string is identical on all four platforms. A 127.0.0.1 bind-probe failed all three native CI jobs (BSD/Windows allow specific-over-wildcard binds); a wildcard probe can steal the port from the server's own bind; a connect-probe consumes a `-s -1` one-off's accept.
- **Sub-ephemeral `free_port()`**: PID-windowed allocation below every platform's ephemeral range, replacing `bind(:0)` — ephemeral allocation let concurrent tests' client *source* ports squat our server ports (3/10 parallel-harness repro in review; now 0/10). Shared by all eight test binaries.
- Client stderr is captured and surfaced in panics (exit status + stderr), and timeout kills drain available output — the next harness failure names its cause.

**Verification:** fmt/clippy clean; riperf3-cli 112/112; workspace 465/465; `bidir_intervals` full binary ×10 parallel and the flaky test ×15: zero failures; adversity-proven (server bound 1 s late is absorbed); all 12 CI checks green including the three native jobs. Test-only change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,10 +80,11 @@ jobs:
   # tcp_connection_info path, winsock) is unverified. These jobs run the real test
   # suite on native runners.
   #
-  # Status (review of the live main ruleset, 2026-06): Native test
-  # (windows-latest) IS a required branch-protection check; macOS remains
-  # informational (reports without blocking; promoted when consistently
-  # clean). fail-fast: false so a failure on one OS doesn't cancel the other.
+  # Status (live main ruleset, 2026-06-10): BOTH matrix legs are required
+  # branch-protection checks — Native test (windows-latest) and Native test
+  # (macos-latest); macOS was promoted from informational after running
+  # consistently clean through the 0.7.2 campaign. fail-fast: false so a
+  # failure on one OS doesn't cancel the other.
   native-test:
     name: Native test (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
@@ -103,7 +104,7 @@ jobs:
 
   # Static/musl build check (#39). riperf3 ships in containers; musl is where
   # libc differences (the libc/nix syscall surface) tend to bite. cargo check
-  # doesn't link, so musl-gcc isn't needed. Informational, like native-test.
+  # doesn't link, so musl-gcc isn't needed. REQUIRED check since 2026-06-10.
   musl-check:
     name: musl check
     runs-on: ubuntu-latest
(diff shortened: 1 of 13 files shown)


therealevanhenry:
Round-1 findings addressed in 02e1f6e (+ main merged in, so the PR diff is test-only again):

- **B1 (the native-CI failure):** port probing is gone entirely. The harness now retries the **client** while it exits with `ConnectionRefused` (10 s window, 100 ms backoff): refused connects never reach `accept()`, so one-off servers are untouched, there is no probe-hold to race the server's bind (B3 dies with it), and the `ConnectionRefused` kind string is identical on all four platforms. Per-attempt `elapsed` keeps the omit timing assertions honest. Adversity-proven: a server deliberately bound 1 s late is absorbed (~1.05 s of refused attempts), output parses clean.
- **B2:** `free_port()` no longer draws from the ephemeral range at all — PID-windowed allocation below 7000+25k, availability-checked **sequentially** per family (a held `v6only=0` wildcard claims the v4 side, so the simultaneous check deadlocked against its own probe — caught locally before push). All eight test binaries now share it, not just the five in r1 scope. Your 3/10 parallel-harness repro now reads **0/10**.
- **Nits:** timeout kills drain available stdout/stderr into the panic; the duplicate #175 commit is resolved by merging main (merge rather than rebase — no force-push policy); the stderr-surfacing half of r1 is retained as you suggested.

Gates: fmt/clippy clean, riperf3-cli 112/112, workspace 465/465, bidir_intervals ×10 parallel 0 failures, udp_bidir ×15 0 failures. The three native CI jobs are the referendum on B1.

Issue #178, opened 2026-06-10
Windows flake: UDP bidir reverse direction carries zero bytes (client receives nothing)


therealevanhenry:
On main run [27288949990](https://github.com/therealevanhenry/riperf3/actions/runs/27288949990) (windows-latest, commit a4acdc7 — the #176 harness fix), `udp_bidir_interval_sums_split_udp_stats` failed `bytes must flow both ways`: a `-u --bidir -t 2 -i 1 -J` run completed end-to-end with **`sum_bidir_reverse.bytes: 0` in every interval and the end block** — the reverse (server→client) flow delivered nothing for the whole run — while the forward direction flowed normally (~46.7 KB/interval at the UDP default rate).

Distinct from #159 (zero *intervals*, both directions carrying bytes — reporter scheduling); this is zero *reverse bytes* with intervals present — a data-path delivery failure. Same Windows-UDP family: plausible suspects are the client-side UDP receive path for the reverse flow (the #80 demux work addressed the server side) and winsock connected-UDP delivery races (#80-class).

Intermittent: the identical tree passed Native test (windows-latest) on PR #176's own CI runs the same day, and this test has passed dozens of Windows runs since #80 landed. Frequency matters because `Native test (windows-latest)` is a **required** check — if this recurs at meaningful rate it will block unrelated PRs and needs either a fix or a documented quarantine decision.

Note: this failure was previously invisible — before #176, harness failures on this test surfaced as `not valid JSON (EOF...)` with stderr nulled. This may have been one of the "flake" occurrences all along, now diagnosable. Investigation home: the mithrandir Windows VM (native winsock build + test, per the riperf3-windows skill).

therealevanhenry:
Escalating from flake to blocking: the rerun (attempt 2) failed with the identical signature — `sum_bidir_reverse` zero throughout, forward normal. Two consecutive main failures, while the same tree passed windows-latest on PR #176's CI ~1h earlier — consistent with a runner-image roll surfacing it at high frequency. `Native test (windows-latest)` is required, so main-targeting PRs are now gated on this. Moving to native repro on the mithrandir VM.

therealevanhenry:
## Root cause found

Native repro attempt: 10/10 PASS on the mithrandir Windows 11 VM (idle) — the bug needs the loaded-runner timing. But code inspection found the mechanism, and it doesn't need a repro:

[...]

PR #179, opened 2026-06-10
fix(stream): UDP receivers survive connection-reset-class ICMP noise (#178)

therealevanhenry:
Hardening for #178 (NOT a fix — see the issue thread: the live mechanism is upstream of recv error handling; the server sends reverse datagrams that never reach the client socket) — the windows-latest `udp_bidir_interval_sums_split_udp_stats` failure (2/2 on main today): a UDP bidir run completing "normally" with `sum_bidir_reverse: 0` in every interval.

**Mechanism:** a connected UDP socket reports ICMP port-unreachable for datagrams it previously *sent* as a recv error (`WSAECONNRESET`/ConnectionReset on Windows; `ECONNREFUSED`/ConnectionRefused on Linux). Both live UDP receive loops — `run_udp_receiver_blocking` (client reverse flow) and `run_udp_server_demux_receiver` (server shared socket, where one event would kill *every* stream's receive) — treated any non-timeout error as terminal and silently stopped receiving. One connect-magic retransmit hitting a closed port in the setup→data transition suffices, which is why it's load/timing-sensitive (GitHub runners reproduce; an idle Windows 11 VM ran it 10/10 green).

**Fix:** the class is transient ICMP noise, not EOF — `continue`. Post-test drain loops keep their `break` (test already over).

**Red-first portable test** (`udp_receiver_survives_connection_reset`): poisons a connected receiver via a send to a just-closed port — Linux surfaces the identical class — then delivers real datagrams from the rebound port. Red on `break` (received nothing), green on `continue`. Linux-only via `cfg_attr(ignore)` since the injection relies on loopback ICMP determinism.

**Faithfulness note:** iperf3 has no handling for this class (the error surfaces loudly as a test failure); `continue` is a deliberate, documented robustness divergence — strictly better than riperf3's prior *silent* zero-data behavior, same precedent as the UDP connect-magic retransmit.

Gates: fmt/clippy clean, 466/466 workspace, msvc/freebsd/darwin cross-checks pass. The windows-latest CI job on this PR (and post-merge main) is the live referendum.

### DIFF
--- a/riperf3/src/stream.rs
+++ b/riperf3/src/stream.rs
@@ -1436,6 +1436,20 @@ pub(crate) fn run_udp_server_demux_receiver(
             {
                 continue
             }
+            // #178: ConnectionReset on a UDP socket is ICMP port-unreachable
+            // noise from something WE sent (Windows surfaces it as
+            // WSAECONNRESET/ConnectionReset; Linux connected-UDP as
+            // ECONNREFUSED/ConnectionRefused) — not EOF. On this
+            // SHARED demux socket a break here would silently end reception
+            // for every stream at once.
+            Err(e)
+                if matches!(
+                    e.kind(),
+                    std::io::ErrorKind::ConnectionReset | std::io::ErrorKind::ConnectionRefused
+                ) =>
+            {
+                continue
+            }
             Err(_) => break,
         }
     }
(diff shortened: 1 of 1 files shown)


therealevanhenry:
Round-2 blocker + both nits addressed in d58dac7: the pin now runs on Windows (the only platform exercising the ConnectionReset arm — the gate comment claiming the injection needs Linux ICMP was factually backwards, as the review showed; winsock latches WSAECONNRESET locally), macOS/FreeBSD stay ignored as explicitly unverified; the assertion pins ≥ half the datagrams (keeps-receiving, not received-once); the demux-arm comment now states the unconnected-socket reality (Linux never surfaces ICMP there without IP_RECVERR — that arm is Windows-live, noted for #180). The required windows-latest check is now the live validation of the test on real winsock.

Issue #184, opened 2026-06-10
Bidir text end block lacks iperf3's per-stream sender/receiver line pairs


therealevanhenry:
## Gap

In UDP `--bidir` **text** mode, iperf3 3.20's end block prints a sender/receiver **line pair per stream**; riperf3 prints a single line per stream with blank datagram columns on the sender row.

iperf3 3.20 ground truth (sandbox fleet, `-u --bidir -t 2`, client text output):

```
[ ID][Role] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  5][TX-C]   0.00-2.00   sec   262 KBytes  1.07 Mbits/sec  0.000 ms  0/30 (0%)  sender
[  5][TX-C]   0.00-2.00   sec   262 KBytes  1.07 Mbits/sec  0.052 ms  0/30 (0%)  receiver
[  7][RX-C]   0.00-2.00   sec   262 KBytes  1.07 Mbits/sec  0.000 ms  0/30 (0%)  sender
[  7][RX-C]   0.00-2.00   sec   262 KBytes  1.07 Mbits/sec  0.040 ms  0/30 (0%)  receiver
```

Note the sender line shows `0.000 ms 0/<sent>` (local sent count, zero jitter/loss) and the paired receiver line carries the peer-measured stats — distinct from the JSON shape, where the sender entry itself carries the peer-measured values (#182).

riperf3's text path (client.rs `print_results_text`, ~1094-1148) emits one `StreamSummary` per stream from local stats only, plus a forward-only graft of server receiver lines (#25 — explicitly skips bidir). So a bidir run shows sender rows with blank jitter/lost columns and no paired receiver rows. TCP bidir likely has the analogous pair-structure gap (retransmit columns) — verify against ground truth when implementing.

Pre-existing (not introduced by #182, which fixed the JSON end block to match iperf3's JSON). Surfaced by #182's review round 1; ground truth captured 2026-06-10.

PR #157, opened 2026-06-09
perf(stream): cumulative-average -b throttle + wire --pacing-timer (#32/#116)

therealevanhenry:
## Summary

Replaces the token-bucket rate limiter with **iperf3's cumulative-average throttle** (`iperf_check_throttle`): a send is green-lit whenever cumulative bytes ≤ elapsed×rate, bounding total overshoot to one in-flight block at any rate. The old bucket granted a `max(rate×0.1, 4×blksize)` burst up front — 512 KiB at TCP's 128 KiB default block — overshooting `-b 1M` by ~2× (#116, verified vs iperf3). High rates self-correct (post-oversleep catch-up sends back-to-back; burstiness ≈ rate × pacing quantum), exactly iperf3's behavior.

`--pacing-timer` is wired end-to-end (#32): CLI → `ClientBuilder::pacing_timer` (additive) → `TestParams` (**always sent**, like iperf3, default 1000 µs) → the server's reverse/bidir sender paces on the client's quantum (absent/0 from older peers → 1000 µs). Wakeups clamp to the quantum, matching iperf3's minimum tick.

**UDP senders untouched**: the blocking/sendmmsg paths use absolute-schedule pacing with no initial-burst defect, and they're the benchmark hot path.

## TDD

Commit 1 is five red tests with the measured defects in the commit message (unit: 655 KB sent vs 295 KB bound at 1 Mbit; integration: 786 KB end-to-end vs 250 KB budget; params/config/CLI wiring all unwired). Commit 2 turns them green; the pre-existing pacing tests (`tcp_bitrate_is_paced` fwd/rev, `tcp_unlimited_is_not_paced`) stay green. 441 tests, clippy/fmt clean.

## Benchmark gate (BEFORE merge — do not merge on CI alone)

This touches the `-b` data path. Gate: standard N=1 matrix (unlimited cells must be drift-free — the limiter is entirely off at `-b 0`) + targeted `-b` accuracy cells vs iperf3 at 1M/200M/5G/10G including the high-rate catch-up check. Results will be posted on this PR.

Closes #32. Closes #116.

Issue #185, opened 2026-06-10
UDP sender bursts a full 32-packet batch then starves at low rate x large blksize (default -b on loopback)


therealevanhenry:
## Symptom

At the default UDP rate (`-b` unset → 1 Mbit/s) on a high-MSS path (loopback: blksize ≈ 32741 from the control MSS), the client sender **bursts a full 32-packet batch (~8.4 Mbit) instantly, then starves for the rest of the run**: one batch consumes ~8.4 s of rate budget, so the pacing sleep parks the sender past the end of a short test.

Observable consequences (verified on loopback, main @ 81b57d6):

- A `-t 1`/`-t 2` default-rate run sends exactly 32 datagrams in one burst — ~4× the `-b` contract averaged over 2 s, then silence (iperf3 paces the same cell smoothly: ~15 packets/s of ~8.9 KB).
- `-u --bidir -O 1` at the default rate reports end-block `bytes: 0, packets: 0` on both streams — ALL traffic lands in the omitted warm-up. Arithmetically correct post-omit accounting, but visually identical to the #182 symptom via a different mechanism (and why #182's omit-path behavior had to be regression-checked at `-b 10M`, where it works: 64 post-omit vs 128 gross).
- Teardown latency: `done` is only observed once per batch loop, so the parked sender holds its blocking thread ~8 s past test end (seen as the ~8 s tail in #178's local instrumented runs).

## Mechanism

`udp_send_loop` (stream.rs ~1190+): `batch_size: u32 = 32` is unconditional — 32 sends between clock checks regardless of rate × blksize, with the batch interval then computed as `per_packet * 32`. Low rate × large datagrams ⇒ batch interval (8.4 s at 1 Mbit/s × 32 KB) far exceeds both the pacing quantum iperf3 uses (~100 ms timer) and short test durations.

## Fix sketch

Scale the batch toward the rate budget — e.g. `batch = clamp(rate_bytes_per_quantum / blksize, 1, 32)` with a ~100 ms quantum (iperf3's pacing timer), keeping 32 as the upper bound for the high-rate fast path. Also lets `done` be observed at quantum granularity instead of 8 s.

Found while probing the `-O` interaction during #182's review round 2.

Issue #163, opened 2026-06-09
UDP -b at moderate rates: throughput far below target with heavy loss (fleet, -b 100M → ~16 Mbit effective, ~30% loss)


therealevanhenry:
Found as a CONTROL while verifying #31 (identical with and without -O, current main): `riperf3 -u -b 100M -t 2` VM-to-VM delivers only ~2.8k packets (~16 Mbit/s effective) with ~30-40% reported loss, while `-b 0` unlimited UDP cells run 30+ Gbit/s clean in the standard matrix. Either the rate-limited UDP send path bursts into receiver-buffer overflow at sub-unlimited rates (the #116 class — the #157 rewrite deliberately did not touch UDP), or the sender never approaches target. Needs an iperf3 same-cell comparison and a look at the blocking UDP sender's pacing at non-zero -b. The compat/bench matrices only exercise `-b 0`, so this cell has no coverage — add one when fixing.

PR #193, opened 2026-06-11
fix: small faithfulness batch — -S strtol base 0, -b rate/burst, pacing nits, blksize bounds, SUM jitter mean (#167, #160, #188, #169)

### LINKED ISSUE #160: Pacing follow-ups from #157 review: unit-suffix parse, -b rate/burst mode, done-recheck after acquire
Three small pre-existing items surfaced by the #157 round-1 review (none introduced by the rewrite):

1. **`--pacing-timer 1K` errors** — iperf3 parses the value with `unit_atoi` (suffixes work); riperf3's arg is a plain integer. Faithfulness nit.
2. **`-b rate/burst` discards the burst count** (client.rs `let (rate, _burst)`) — iperf3's burst mode sends N blocks per green light; riperf3 parses and drops it. The cumulative-average limiter could honor it (allow `burst` blocks past green before sleeping).
3. **Sender can write one block after `done`** — `acquire()` can sleep past test end and then send without re-checking the flag (worst case one extra uncounted block + shutdown delayed by the residual sleep at very low `-b`). Modern iperf3's worker re-checks before sending. Cheap fix: re-check `done` after `acquire` returns in `run_tcp_sender`.

### LINKED ISSUE #167: -S/--tos rejects hex/octal values iperf3 accepts (strtol base 0)
Found incidentally in the #153 round-3 review (pre-existing): `-S 0x20` is rejected by clap (`tos: Option<i32>` plain integer parse) while iperf3 parses TOS with strtol base 0 (hex/octal/decimal per its man page). Faithful fix: a custom value parser mirroring strtol base-0 semantics (the `--dscp` parser already does dec/hex/octal — reuse its number-parsing arm). Add accepted/rejected wiring tests both ways.

### LINKED ISSUE #169: END-block summary jitter should be the mean across streams, not the max
## Divergence

iperf3's END-block summary jitter is the **mean** across streams: it accumulates `avg_jitter += sp->jitter` per stream (`iperf_api.c:4249`) and divides by the stream count (`avg_jitter /= num_streams`, `iperf_api.c:4490`). The same mean-per-direction logic applies in bidir.

riperf3's `sum_summaries` aggregation uses **MAX** of the per-stream jitters for the summary jitter value.

With `-P 1` (the common case) the two are identical, which is why the compat matrix doesn't catch it. With `-P > 1` UDP the `[SUM]` jitter line and the `-J` `end.sum.jitter_ms` diverge from iperf3.

## Fix sketch

In the summary aggregation, accumulate `jitter_sum` and divide by the number of UDP receiving streams in that direction (mirroring the per-direction interval mean already landed for #142 in PR #166).

## Notes

- Found during PR #166 review (round 1) while fixing the interval-level jitter mean (#142) — the END block has the same bug one layer up.
- Faithfulness class: report-model parity with iperf3.

### LINKED ISSUE #188: Reject block size 0 (-l 0) at config like iperf3 (IEBLOCKSIZE); currently panics/degenerate
## Gap

riperf3 accepts `-l 0` (UDP/TCP block size 0) where iperf3 rejects it at config time (`IEBLOCKSIZE`). Two symptoms:

- `riperf3 -c host -u -l 0 -t 1` panics a worker thread at `stream.rs` (UDP header write into a zero-length buffer: "range end index 4 out of range for slice of length 0").
- `-l 0` TCP runs a degenerate zero-byte test instead of erroring.

A negotiated `len: 0` from a peer reaches the server the same way. #184 added `.max(1)` guards at the report/packet-count arithmetic (which closed a divide-by-zero DoS in the results exchange), but that papers over the report path — the data path's zero-length buffer is still wrong, and the real faithful behavior is to **reject blksize 0 at config** like iperf3.

## Fix

Validate block size > 0 in the client/server builders (and on received params), erroring like iperf3's `IEBLOCKSIZE` rather than running a degenerate test. Iperf3's bound is `1 <= blksize <= MAX_BLOCKSIZE`; riperf3 already enforces the upper bound (MAX_UDP_BLKSIZE) — this is the missing lower bound.

Found in #184 review round 2. The `.max(1)` guards from #184 can then become assertions/be removed once config rejects 0.

therealevanhenry:
## Small faithfulness batch (0.7.3, batch 1 of the campaign)

Four small parity fixes, bundled per the PR #166 precedent. Every behavioral claim was audited against the iperf3 source (`iperf_api.c` @ 8701e52) before implementation.

### #169 — END-block SUM jitter is the mean, not the max
iperf3: `avg_jitter += sp->jitter` per stream of the direction, then `avg_jitter /= test->num_streams` (iperf_api.c:4249, 4490). `sum_summaries` used worst-case; identical at `-P 1`, divergent at `-P ≥ 2` UDP (text `[SUM]` + `-J end.sum.jitter_ms`). Same fix the interval layer got in #166 (#142).

### #167 — `-S/--tos` accepts strtol-base-0 forms
`-S 0x20` / `-S 020` now parse (shared numeric arm with `--dscp`), with iperf3's IEBADTOS 0-255 range check (iperf_api.c:1504-1511) at string parse and at `build()` for the lib setter path. CLI carries the raw string through `ClientBuilder::tos_str` (lib stays clap-free).

### #160 — pacing follow-ups from the #157 review
- **`-b rate/burst` honored**: the limiter green-lights `burst` blocks per throttle check (iperf3's multisend loop, iperf_api.c:2193-2213); the count rides the param exchange (`"burst"`, sent only when nonzero, iperf_api.c:2461) so reverse/bidir servers batch too; present-burst range is 1..=1000 per IEBURST — `/0` is an error, distinct from no slash.
- **`--pacing-timer 1K` works**: iperf3 parses it with `unit_atoi` (1024-based, iperf_api.c:1780); the old plain-u32 clap parser rejected suffixes. The i32::MAX wire cap moved into `pacing_timer_str`.
- **`done` re-checked after the throttle sleep**: at very low `-b` the green-light sleep can outlast the test; the sender wrote one extra block past `done` (red test observed 2 blocks where 1 was correct).

### #188 — block size bounds at config, like iperf3
iperf3's semantics (iperf_api.c:1926-1942) are subtler than the issue stated: `blksize == 0` maps to the **default** (UDP: dynamic-from-MSS), then TCP validates `1..=MAX_BLOCKSIZE` (IEBLOCKSIZE) and UDP `MIN_UDP_BLOCKSIZE..=MAX_UDP_BLOCKSIZE` (IEUDPBLOCKSIZE). Implemented exactly that at `ClientBuilder::build()`, and the same bounds on a **negotiated** `len` in the server's `from_params` — closing the `-l 0` UDP header-write panic, a negative-len `as usize` wrap (multi-EiB alloc), and the oversized-alloc DoS from hostile peers. No SERVER_ERROR state on rejection: iperf3 clients follow that with an errno exchange we don't speak, and real clients never send these values — the server drops the test instead. The #184 `.max(1)` guards stay as defense.

### Wire note
`TestParams` gains an optional `"burst"` field, serialized only when nonzero — exactly iperf3's `if (test->settings->burst) cJSON_AddNumber(...)`, so the param JSON stays byte-compatible for every flag combination that exists today.

### Verification
- TDD: red observed for the limiter burst window, the done re-check, `parse_bitrate` burst range, `from_params` bounds, and the jitter mean; all green after.
- Full workspace suite green; clippy/fmt clean; xcheck green on all 7 targets.
- Smoke compat-matrix drift gate to run post-merge per campaign process.

Fixes #167. Fixes #169. Fixes #188. Closes #160.

### DIFF
--- a/riperf3-cli/src/cli.rs
+++ b/riperf3-cli/src/cli.rs
@@ -161,9 +161,11 @@ pub struct Cli {
     #[arg(short = 'b', long, value_name = "rate[/burst]")]
     pub bitrate: Option<String>,
 
-    /// Set the IP type of service (0-255)
+    /// Set the IP type of service (0-255; decimal, 0x hex, or 0 octal)
+    // String, parsed by the builder's `tos_str` with iperf3's strtol-base-0
+    // semantics + IEBADTOS range check (#167).
     #[arg(short = 'S', long, value_name = "tos")]
-    pub tos: Option<i32>,
+    pub tos: Option<String>,
 
     /// Omit the first N seconds of the test
     #[arg(short = 'O', long, value_name = "secs")]
@@ -202,10 +204,11 @@ pub struct Cli {
     pub cport: Option<u16>,
 
     /// Set the server timing for pacing in microseconds (deprecated)
-    // Capped at i32::MAX: the wire TestParams field is i32 (a larger u32
-    // would wrap negative on the wire — review r1).
-    #[arg(long, value_name = "usec", value_parser = clap::value_parser!(u32).range(..=i32::MAX as i64))]
-    pub pacing_timer: Option<u32>,
+    // String, parsed by the builder's `pacing_timer_str`: iperf3 reads it
+    // with unit_atoi, so KMG suffixes (1024-based) work; the i32::MAX wire
+    // cap (review r1) is enforced there too (#160).
+    #[arg(long, value_name = "usec")]
+    pub pacing_timer: Option<String>,
 
     /// Only use IPv4
     #[arg(short = '4', long, conflicts_with = "version6")]
@@ -487,8 +490,8 @@ impl Cli {
         if let Some(t) = self.time {
             builder = builder.duration(t);
         }
-        if let Some(us) = self.pacing_timer {
-            builder = builder.pacing_timer(us);
+        if let Some(ref s) = self.pacing_timer {
+            builder = builder.pacing_timer_str(s)?;
         }
         if let Some(ref s) = self.bytes {
             builder = builder.bytes_str(s)?;
@@ -523,8 +526,8 @@ impl Cli {
         if let Some(ref s) = self.bitrate {
             builder = builder.bandwidth_str(s)?;
         }
-        if let Some(tos) = self.tos {
-            builder = builder.tos(tos);
+        if let Some(ref s) = self.tos {
+            builder = builder.tos_str(s)?;
         }
         if let Some(o) = self.omit {
             builder = builder.omit(o);
@@ -1155,6 +1158,25 @@ mod cli_tests {
             assert_eq!(c, expected_client("host").tos(16).build().unwrap());
         }
 
+        // #167: -S takes iperf3's strtol-base-0 forms (hex/octal), and bad
+        // values fail the build with IEBADTOS wording instead of parsing.
+        #[test]
+        fn tos_flag_accepts_hex_and_octal() {
+            let cli = Cli::parse_from(["riperf3", "-c", "host", "-S", "0x20"]);
+            let c = build_client_from_cli(&cli);
+            assert_eq!(c, expected_client("host").tos(0x20).build().unwrap());
+            let cli = Cli::parse_from(["riperf3", "-c", "host", "-S", "020"]);
+            let c = build_client_from_cli(&cli);
+            assert_eq!(c, expected_client("host").tos(16).build().unwrap());
+        }
+
+        #[test]
+        fn tos_flag_rejects_out_of_range() {
+            let cli = Cli::parse_from(["riperf3", "-c", "host", "-S", "256"]);
+            let err = cli.build_client().unwrap_err().to_string();
+            assert!(err.contains("bad TOS value"), "got: {err}");
+        }
+
         #[test]
         fn omit_flag_wired() {
             let cli = Cli::parse_from(["riperf3", "-c", "host", "-O", "3"]);
@@ -1636,6 +1658,14 @@ mod cli_tests {
         }
 
         // #32: --pacing-timer was parsed but never wired — a silent no-op.
+        #[test]
+        fn pacing_timer_accepts_kmg_suffix() {
+            // #160: iperf3 parses --pacing-timer with unit_atoi (1024-based).
+            let cli = Cli::parse_from(["riperf3", "-c", "h", "--pacing-timer", "1K"]);
+            let c = build_client_from_cli(&cli);
+            assert_eq!(c, expected_client("h").pacing_timer(1024).build().unwrap());
+        }
+
         #[test]
         fn pacing_timer_wired() {
             let cli = Cli::parse_from(["riperf3", "-c", "h", "--pacing-timer", "500"]);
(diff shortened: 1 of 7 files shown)


therealevanhenry:
## Review round 1 → resolved (1848c29)

- **Blocker (UDP burst)**: implemented, not scoped away. An explicit burst now IS the UDP batch — `udp_pacing_batch` returns it, overriding the #185 quantum heuristic and the `max_batch` ceiling (iperf3's only cap is MAX_BURST=1000, enforced at parse) — threaded through both `run_udp_sender_sendmmsg` variants, `run_udp_sender_blocking`, and `run_udp_server_demux_sender` on both roles. Unit-pinned in `udp_pacing_batch_scales_with_rate` (burst 5 → 5; burst 1000 > ceiling honored; burst at `-b 0` honored, matching iperf3's `if (burst) multisend = burst` precedence).
- **Nit 2 (pacing-timer wire cap)**: `pacing_timer_str_enforces_i32_wire_cap` — `3G` rejected, `2147483647` ok, `2147483648` rejected.
- **Nit 3 (IEBADTOS wording)**: `parse_tos` raises the IEBADTOS text for unparsable input too; the deliberate stricter-than-strtol full-string parse (rejecting `32abc`/`08`/`0x`) is recorded in a comment per note 4.

Full workspace green, clippy/fmt clean, xcheck green on all 7 targets.

Flake note: the `Test (Linux)` failure on the first run (`udp_bidir_end_block_pairs_each_stream` timeout) reproduced at 1/20 under the faithful 2-core repro (`taskset -c 0,1`, `--test-threads=2`) on this branch; base-ref attribution loop running. Rerun passed; tracking separately from this PR if base shows the same rate.

therealevanhenry:
Flake attribution complete: base ref (main @ a92ebd0) also fails the faithful 2-core loop at 1/20 with the identical timeout shape — pre-existing, not from this PR. Filed as a tracking issue; CI rerun is green.

therealevanhenry:
## Review round 2 → resolved (19c0b28)

- **Blocker 1 (uninterruptible green-light debt)**: both UDP absolute-schedule pacing blocks now share `pace_until()` — 100 ms slices re-checking `done` and the `-t` deadline, preserving the 20 µs margin + spin-to-line. `RateLimiter::acquire` takes `done` and returns early without accounting; the caller's post-acquire re-check breaks. Pinned by a wall-clock bound on the existing virtual-time TCP test (~5 s exit, not the 10 s green-light instant) and a new real-time UDP test (join < 1.5 s against a 2 s batch debt).
- **Blocker 2 (unbounded wire burst)**: `from_params` rejects `burst > 1000` per the len posture (no real iperf3 client can produce it — its own IEBURST parse caps at MAX_BURST); non-positive = unset. Test mirrors `from_params_validates_len_bounds`.
- **n3 (burst=1 SNDBUF shrink)**: that's exactly the #163 mechanism — the grow-only `configure_udp_sender` fix lands in that PR (next in the campaign queue) rather than riding here.
- **n4**: `build_params_sends_burst_only_when_set` pins the wire serialization.
- **n5**: noted as cosmetic; leaving the UDP-arm-first order (absurd-input wording only).
- **n6**: that's #151 — PR #194 (in flight) routes every builder error through the `riperf3: error - {Display}` printer.

297 lib tests green, full workspace green, clippy/fmt clean, xcheck 7/7.

Issue #161, opened 2026-06-09
snd_wnd: riperf3 emits 0 where iperf3 emits -1 (macOS) / real value (Linux)


therealevanhenry:
From the #152 round-2 review (pre-existing pattern, newly observable on macOS now that Retr/Cwnd are un-suppressed there):

riperf3 emits interval `snd_wnd: 0` alongside the TCP detail (reporter.rs `snd_wnd: snd_cwnd.map(|_| 0u64)`). iperf3's macOS `get_snd_wnd()` returns **-1** (its `HAVE_TCP_INFO_SND_WND` probe checks `struct tcp_info`, which macOS userland doesn't expose, so the APPLE `tcpConnInfo.tcpi_snd_wnd` branch is dead) — and on Linux iperf3 emits the real value where available. Ironically #152's hand-rolled binding DOES read the real `tcpi_snd_wnd` into the snapshot, where it sits `#[allow(dead_code)]`.

Fix direction: per-platform faithfulness — match iperf3's actual emitted value per OS (-1 on macOS, real value where iperf3 has it, 0 where iperf3 emits 0). Small, but needs an iperf3-source audit per platform first.

therealevanhenry:
Scoped down by PR #194 (audit of iperf3's per-OS reality in `tcp_info.c` `get_snd_wnd` + `configure.ac`):

- **Linux**: iperf3's `HAVE_TCP_INFO_SND_WND` probe prefers `linux/tcp.h`, so Linux builds emit the **real** `tcpi_snd_wnd`. riperf3 now does too (UAPI `tcp_info` mirror; libc's glibc-shaped struct truncates before the field). Same mirror unlocks `tcpi_reord_seen`, so `reorder` is now live on Linux as well (iperf3 `get_reorder`, 4.19+).
- **FreeBSD**: field exists in `netinet/tcp.h` → probe alive → raw value. riperf3 now forwards the already-captured value.
- **macOS**: no `struct tcp_info` → probe dead → every macOS iperf3 emits **-1** (the APPLE `tcpConnInfo.tcpi_snd_wnd` branch is unreachable). riperf3 pins 0 there for now: carrying `-1` requires flipping the pub report fields `snd_wnd`/`max_snd_wnd` (and `reorder` for the non-Linux `-1`) from unsigned to `i64`, which is a SemVer break the CI gate rightly rejects for a patch release.

[...]

Issue #96, opened 2026-06-04
macOS: report real sender retransmits (Retr/Cwnd) via tcpi_txretransmitpackets


therealevanhenry:
Follow-up to #40 (addressed in #95).

## Background
#95 stops macOS from advertising retransmit info (`has_retransmit_info()` no longer returns true on macOS), because `tcp_info::get_tcp_info` can only hard-code `total_retransmits: 0` there — so the interval report no longer shows a misleading `Retr 0`. But the Retr and Cwnd columns are gated together (iperf3 couples them), so **this also suppresses the Cwnd column on macOS**.

## The divergence
This is *less* faithful than iperf3, which **does** show Retr+Cwnd on macOS with real data:
- `has_tcpinfo_retransmits()` returns 1 on macOS (esnet/iperf `src/tcp_info.c`, `__APPLE__ && TCP_CONNECTION_INFO`).
- `get_total_retransmits()` reads `tcpConnInfo.tcpi_txretransmitpackets` (a real sender retransmit packet count).

riperf3 can't read that today: Rust's `libc` 0.2.x `tcp_connection_info` binding exposes `tcpi_txretransmitbytes` (sender *bytes*) and `tcpi_rxretransmitpackets` (receiver packets) but **not** `tcpi_txretransmitpackets` (the sender packet count iperf3 uses).

## Faithful fix
- Add `tcpi_txretransmitpackets` via a hand-rolled `tcp_connection_info` binding (or upstream a fix to the `libc` crate), then populate `TcpInfoSnapshot.total_retransmits` from it on macOS.
- Re-add macOS to `has_retransmit_info()` so the real Retr+Cwnd columns return.
- Verify on the `macos-latest` native CI job.

Deferred from #95 (the 0.6.3 fidelity batch) as too platform-specific for that patch. See the rationale comment on `tcp_info::has_retransmit_info`.

Issue #15, opened 2026-05-29
Client -B (local bind address) is ignored


therealevanhenry:
## Summary

On the **client**, `-B`/`--bind` (local bind address) is parsed and stored but never used — the connection doesn't bind to it. Surfaced during cold review of PR #13.

## Evidence

- `Client.bind_address` is set from the CLI (`main.rs`) and the builder, but neither `net::tcp_connect` nor the UDP client path consults it — both pass `None` for the local bind. (PR #13 wires `cport` but not `bind_address`.)

## Impact

`riperf3 -c <host> -B <local-addr>` silently does nothing — the OS picks the source address. Combined with #10, the client also doesn't validate `-B`'s family against `-4`/`-6` (the #12 validation only covers the server).

## Expected

Bind the client's data/control sockets to the given local address (and validate its family against `-4`/`-6`, mirroring #12 on the server).

## Notes

Pre-existing; out of scope for #13 (which fixes the *host*-family selection). Tracking so the `-B` story is finished on both sides.

PR #194, opened 2026-06-11
fix: iperf3 error shape for CLI errors; live tcpi_snd_wnd/reord_seen on Linux+FreeBSD (#151, #161)

### LINKED ISSUE #151: CLI prints top-level errors in Rust Debug form, not iperf3's 'iperf3: error - ...' format
Surfaced during the #148 round-2 review (pre-existing, CLI-wide).

`main()` returns `Result<(), Box<dyn Error>>`, so any propagated error prints as Rust's Debug rendering, e.g.:

```
Error: Io(Os { code: 19, kind: Uncategorized, message: "No such device" })
```

Real iperf3 prints `iperf3: error - <human text>: <errno text>` and exits 1 (e.g. `iperf3: error - unable to connect to server - server may have stopped running or use a different port, firewall issue, etc.: No such device`).

For a drop-in, scripts that grep stderr for `iperf3: error` (or just expect readable text) see a very different shape. Fix direction: a small top-level error printer in main (catch the Err, print `riperf3: error - {Display}` to stderr, exit 1), plus Display impls that read like iperf3's IE* strings where we already mirror them. Audit the existing rejection messages (#65/#100 paths) which already embed iperf3 wording — those go through the same Box<dyn Error> path and would benefit uniformly.

Candidate for 0.7.2's #45 error-policy PR or its own small PR.

therealevanhenry:
### #151 — CLI errors print in iperf3's `error - ...` shape
`main()` returned `Result<(), Box<dyn Error>>`, so every propagated error rendered in Rust Debug form (`Error: Io(Os { code: 111, ... })`). iperf3 prints `iperf3: error - <human text>` and exits 1 (`iperf_errexit`); scripts grep for that shape. `main` now catches and prints `riperf3: error - {Display}`, exit code 1. The control-connect failure also carries iperf3's IECONNECT wording (`unable to connect to server - server may have stopped running or use a different port, firewall issue, etc.: ...`), with the `io::ErrorKind` preserved (the CLI test harness's refused-retry keys off it).

### #161 — live `tcpi_snd_wnd` / `tcpi_reord_seen` where iperf3 has them
Audited iperf3's per-OS reality first: its `HAVE_TCP_INFO_SND_WND` configure probe prefers `linux/tcp.h`, so **Linux** iperf3 emits the real peer-advertised window in `-J` interval `snd_wnd` and end `max_snd_wnd`; **FreeBSD**'s `tcp_info` has the field too (raw); **macOS** has no `struct tcp_info`, the probe is dead, and `get_snd_wnd` returns `-1` everywhere. Same shape for `get_reorder`: Linux reads `tcpi_reord_seen` (4.19+), everywhere else `-1`.

- **Linux**: libc's glibc-shaped `tcp_info` truncates after `tcpi_total_retrans`, so riperf3 pinned `snd_wnd`/`reorder` to 0. New hand-rolled UAPI `tcp_info` mirror through `tcpi_rcv_wnd` (the #96 macOS-binding tradition): `offset_of!` const asserts pin the ABI (`tcpi_snd_wnd` @ 228, `tcpi_reord_seen` @ 220), zeroed init + returned-length gates keep pre-5.4/pre-4.19 kernels at 0 exactly like a short kernel copy.
- **FreeBSD**: the already-captured `tcpi_snd_wnd` now flows to the report (was `#[allow(dead_code)]`).
- **macOS**: pinned 0 with rationale — emitting the live xnu value would diverge from every macOS iperf3 build (-1); the faithful `-1` needs `i64` report fields, a semver break deferred to 0.8.0 (comment incoming on #161, which stays open for that).
- `StreamExtremes`/`TcpEndExtras` gain `max_snd_wnd` (both `#[non_exhaustive]`, additive — SemVer-clean).

### Verification
- TDD: `error_format` CLI test red (Debug shape) → green; `linux_snapshot_reads_real_snd_wnd` red (always 0) → green.
- Full workspace green, clippy/fmt clean, xcheck green on all 7 targets.
- Smoke compat-matrix drift gate post-merge per campaign process.

Fixes #151. (#161 stays open, scoped down to the 0.8.0 `-1` sentinel item.)

### DIFF
--- a/riperf3-cli/src/main.rs
+++ b/riperf3-cli/src/main.rs
@@ -7,7 +7,21 @@ use log4rs::encode::pattern::PatternEncoder;
 mod cli;
 use cli::Cli;
 
-fn main() -> std::result::Result<(), Box<dyn std::error::Error>> {
+fn main() -> std::process::ExitCode {
+    match run() {
+        Ok(()) => std::process::ExitCode::SUCCESS,
+        Err(e) => {
+            // iperf3's iperf_errexit shape ("iperf3: error - <text>", exit 1)
+            // instead of Rust's Debug rendering; ours carries the actual
+            // binary name. The Display strings already mirror iperf3's IE*
+            // wording where riperf3 implements the same rejections (#151).
+            eprintln!("riperf3: error - {e}");
+            std::process::ExitCode::FAILURE
+        }
+    }
+}
+
+fn run() -> std::result::Result<(), Box<dyn std::error::Error>> {
     let cli = Cli::parse();
 
     configure_log4rs(cli.debug.unwrap_or(0));
--- a/riperf3-cli/tests/common/mod.rs
+++ b/riperf3-cli/tests/common/mod.rs
@@ -112,7 +112,15 @@ pub fn run_client(args: &[&str], timeout: Duration, who: &str) -> ClientRun {
             .unwrap();
 
         if !status.success()
-            && (stderr.contains("ConnectionRefused") || stderr.contains("Connection refused"))
+            && (stderr.contains("ConnectionRefused")
+                || stderr.contains("Connection refused")
+                // Windows WSAECONNREFUSED renders as "No connection could be
+                // made because the target machine actively refused it.
+                // (os error 10061)" — neither other token matches, and this
+                // retry is the ONLY server-readiness mechanism (the doc
+                // above): dropping it reopens the bind-race flake on the
+                // required windows-latest check (review r2).
+                || stderr.contains("(os error 10061)"))
             && Instant::now() < retry_deadline
         {
             // Server not listening yet — give it a beat and go again.
--- a/riperf3-cli/tests/common/mod.rs
+++ b/riperf3-cli/tests/common/mod.rs
@@ -112,7 +112,7 @@ pub fn run_client(args: &[&str], timeout: Duration, who: &str) -> ClientRun {
             .unwrap();
 
         if !status.success()
-            && stderr.contains("ConnectionRefused")
+            && (stderr.contains("ConnectionRefused") || stderr.contains("Connection refused"))
             && Instant::now() < retry_deadline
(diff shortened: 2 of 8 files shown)


therealevanhenry:
## Review round 1 → resolved (eebf7db)

- **Blocker 1 (22-space literal)**: collapsed; the `error_format` test now pins the FULL canonical IECONNECT sentence (observed red first, exactly per finding 6).
- **Blocker 2 (--connect-timeout bypass)**: `ConnectionTimeout` folds into the IECONNECT sentence with strerror(ETIMEDOUT)'s text, kind `TimedOut` preserved. Scoped to dial-class errors — the first attempt wrapped everything and broke the pinned #15 family-conflict Protocol classification (caught by `client_ip_version_conflicts_with_literal_host`).
- **n3 (-B bind path)**: recorded deviation in the wrap-site comment (shares the Protocol variant with config validations; reclassifying needs a net.rs error split).
- **n4 (os-error suffix)**: explicit recorded decision in the comment — substring matchers survive, strerror text varies by platform/locale anyway.
- **n5/n7**: stale comments fixed; kernel ≥ 5.4 prerequisite documented on the test.
- **n8/n9**: filed as #198 (error-sink parity, 0.7.3) and #199 (omit-boundary extremes reset, 0.7.3).

Full workspace green, clippy/fmt clean.

therealevanhenry:
## Review round 2 → resolved

- **Blocker (Windows refused-retry dead)**: the matcher now also takes `(os error 10061)` with a comment explaining all three tokens and why the retry is load-bearing on windows-latest. (PR #196's shared `refused()` got the identical fix — the two converge when both merge.)
- **n2**: resolve_host's escaping Protocol path added to the deviation ledger comment alongside the `-B` case.
- **n3**: the glibc-vs-BSD strerror nuance recorded in the comment.

error_format green; fmt clean.

PR #196, opened 2026-06-11
test: shared riperf3-test-support crate; retire fixed bind sleeps (#192, #177)

### LINKED ISSUE #81: `-s -D` daemon-mode server accepts connections but never serves; any client hangs
## Summary

A daemonized server (`riperf3 -s -D`) opens its listening socket but never runs the test. Any client — riperf3 **or** iperf3 — connects and then hangs indefinitely. A non-daemon server (`riperf3 -s`, foreground) works normally.

Found while building the throughput benchmark harness (`scripts/bench.sh`), which now shell-backgrounds the server with `nohup` instead of `-D` as a workaround.

## Characterization

All client runs `-t 3` against a fresh server, on an isolated VM pair (Debian 13, Linux 6.12.90, x86_64):

| client → server | result |
|---|---|
| riperf3 → riperf3 `-s -D -1` | **hang** |
| iperf3  → riperf3 `-s -D -1` | **hang** (client-agnostic) |
| riperf3 → iperf3  `-s -D -1` | OK (riperf3 client is fine) |
| riperf3 → riperf3 `-s -D` (no `-1`) | **hang** (not about `-1`) |
| riperf3 → riperf3 `-s` (foreground) | OK (control) |

The fault is the **riperf3 daemon-mode server**, independent of the client and of `-1`. The socket reaches `LISTEN` (confirmed via `ss`), the client's TCP connect succeeds, but the test never proceeds.

## Likely root cause

Classic signature of **forking to daemonize _after_ the tokio runtime is built**: `fork()` carries only the calling thread into the child, so the daemon child has no runtime worker threads to drive the accept/serve loop. Fix: daemonize (fork + `setsid`) **before** constructing the runtime / entering async, or re-establish the runtime in the child.

## Repro

    # server
    riperf3 -s -D -p 5201
    ss -ltn | grep 5201          # shows LISTEN
    # client
    riperf3 -c <server> -p 5201 -t 5     # hangs

## Notes / impact

- Appears **pre-existing** — 0.6.0 did not touch the daemonize path; latent because no test or interop path exercised `-D`.
- The README Platform-Support matrix lists `-D` as supported (Linux), so this is a documented-but-broken flag — a fidelity gap vs iperf3, whose `-D` works.

Environment: riperf3 0.6.0, Debian 13 (Trixie) / Linux 6.12.90-cloud, x86_64.

### LINKED ISSUE #177: Test harness: udp_demux and daemon still gate inline clients on fixed sleeps
Follow-up from PR #176 (r2 review nit N2): `udp_demux.rs` and `daemon.rs` still use fixed ~300 ms sleeps before their inline clients — the same flake class #176 eliminated elsewhere, at lower stakes (their clients assert success, but `udp_demux` nulls client stderr, so a refused client fails semi-blind).

`udp_demux` isn't a drop-in for `common::run_client_ok` (it drains stdout concurrently), so it needs a variant — e.g. a `run_client` flavor that streams stdout, or a refused-retry wrapper around its existing drain loop. `daemon.rs` may be closer to drop-in.

Both files already share `common::free_port()` (landed in #176).

### LINKED ISSUE #192: Consolidate test helpers (free_port, udp_serial, CLI harness) into a shared test-support crate
## Improvement: a shared dev-only test-support crate

`free_port` (the #176 sub-ephemeral PID-windowed allocator) is now copy-pasted in **three** places:
- `riperf3-cli/tests/common/mod.rs`
- `riperf3/tests/common/mod.rs` (added with the #178 lib regression test)
- and inline-derived again in `riperf3/tests/integration.rs` (a plain static counter from 15201, no bind check — already a drift instance the #186 review flagged).

Each copy must be re-patched independently for the next port-collision fix, and the copies have already started to diverge (one carries an abridged comment; integration.rs's scheme can even collide with the PID-windowed one ~1/250). The new per-binary `udp_serial` lock (#191) is similarly destined to be re-implemented wherever the next handshake-heavy UDP test binary lands.

## Proposal

A `riperf3-test-support` dev-dependency crate (or a `[dev-dependencies]`-shared module) housing:
- `free_port()` — single source of the #176 allocator;
- `udp_serial()` — the per-binary UDP-test serialization guard (#191), so handshake-heavy UDP tests don't starve each other on a 2-core runner;
- the `run_client` / `ChildGuard` CLI harness helpers currently duplicated between `riperf3-cli/tests/common` and ad-hoc copies.

Scope: dev/test only — no product or public-API change, no SemVer impact. Patch-class.

Background: surfaced during the #182/#184/#185 follow-up campaign (2026-06-10); the #186 review explicitly flagged the `free_port` duplication.

therealevanhenry:
## Test infra: shared `riperf3-test-support` crate (#192) + fixed-sleep retirement (#177)

Dev/test only — no product or public-API change, no SemVer impact.

### #192 — single source for the duplicated harness helpers
New `riperf3-test-support` workspace member (`publish = false`, dev-dependency of both crates):
- `free_port()` — the #176 PID-windowed sub-ephemeral allocator. Was byte-identical in two `common/mod.rs` files and divergently re-derived in `riperf3/tests/integration.rs` as a bare 15201+ counter with no collision probe (the drift instance the #186 review flagged — it could land inside another binary's PID window ~1/250).
- `udp_serial()` — the #191 per-binary UDP serialization lock. Statics in the rlib are per-linked-binary, so the lock semantics are unchanged (documented in the crate).
- `run_client_with`/`run_client_ok_with`/`ClientRun` — the #176 refused-retry runner. `CARGO_BIN_EXE_riperf3` only exists while compiling riperf3-cli's own tests, so the support fns take the bin path and `common/mod.rs` keeps thin same-signature wrappers — **no test call sites change shape**.
- `ChildGuard` + `wait_bounded` — were private copies in `udp_demux.rs`/`end_block_text.rs`.

### #177 — no more fixed bind sleeps
- `udp_demux.rs`: the 300 ms pre-client sleep is replaced by a bounded refused-retry around its bespoke concurrent-drain client; stderr is now captured (was nulled — a refused client failed semi-blind, the old `not valid JSON (EOF...)` riddle class).
- `daemon.rs`: the 300 ms sleep + hand-rolled bounded wait are replaced by `run_client` (hard timeout still bounds the pre-fix #81 never-serves hang, with the #81 context in the `who` string).
- The shared `refused()` classifier matches both the current Debug rendering and #194's iperf3-shaped message, so this is merge-order-proof.

### Verification
Full workspace green; clippy/fmt clean. The touched binaries (`daemon`, `udp_demux`, `end_block_text`) pass under the faithful 2-core condition. Test-only change.

Fixes #192. Fixes #177.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -800,6 +800,7 @@ dependencies = [
  "log",
  "nix",
  "rand 0.9.2",
+ "riperf3-test-support",
  "rpassword",
  "rsa",
  "serde",
@@ -822,10 +823,15 @@ dependencies = [
  "log4rs",
  "nix",
  "riperf3",
+ "riperf3-test-support",
  "serde_json",
  "thiserror 2.0.3",
  "tokio",
 ]
+
+[[package]]
+name = "riperf3-test-support"
+version = "0.7.2"
 
 [[package]]
 name = "rpassword"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ resolver = "2"
 members = [
     "riperf3",
     "riperf3-cli",
+    "riperf3-test-support",
 ]
 
 [workspace.package]
--- a/riperf3-cli/Cargo.toml
+++ b/riperf3-cli/Cargo.toml
@@ -29,6 +29,7 @@ nix.workspace = true
 
 [dev-dependencies]
 serde_json.workspace = true
+riperf3-test-support = { path = "../riperf3-test-support" }
 
 [[bin]]
 name = "riperf3"
--- a/riperf3-cli/tests/bidir_intervals.rs
+++ b/riperf3-cli/tests/bidir_intervals.rs
@@ -17,14 +17,8 @@ fn free_port() -> u16 {
     common::free_port()
 }
 
-/// Kills the wrapped child on drop so a spawned server is reaped on panic.
-struct ChildGuard(std::process::Child);
-impl Drop for ChildGuard {
-    fn drop(&mut self) {
-        let _ = self.0.kill();
-        let _ = self.0.wait();
-    }
-}
+// Reaper guard shared via riperf3-test-support (#192).
+use common::ChildGuard;
 
 /// Run the client to completion (with refused-retry) and return its stdout.
 fn run_capturing(args: &[&str], timeout: Duration, who: &str) -> String {
--- a/riperf3-cli/tests/byte_limit_summary.rs
+++ b/riperf3-cli/tests/byte_limit_summary.rs
@@ -17,14 +17,8 @@ fn free_port() -> u16 {
     common::free_port()
 }
 
-/// Kills the wrapped child on drop so a spawned server is reaped on panic.
-struct ChildGuard(std::process::Child);
-impl Drop for ChildGuard {
-    fn drop(&mut self) {
-        let _ = self.0.kill();
-        let _ = self.0.wait();
-    }
-}
+// Reaper guard shared via riperf3-test-support (#192).
+use common::ChildGuard;
 
 /// Run the client to completion (with refused-retry) and return its stdout.
 fn run_capturing(args: &[&str], timeout: Duration, who: &str) -> String {
--- a/riperf3-cli/tests/common/mod.rs
+++ b/riperf3-cli/tests/common/mod.rs
@@ -1,143 +1,25 @@
 //! Shared CLI-test helpers (`mod common;` per test binary).
+//!
+//! The implementations live in the dev-only `riperf3-test-support` crate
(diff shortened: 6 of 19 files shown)


therealevanhenry:
## Review round 1 → resolved

- **Blocker**: consolidation completed — all 7 remaining private `ChildGuard` copies migrated; json_stream's panicking `wait_bounded` is now a thin shim over the shared one. `Fixes #192` stands.
- **n2**: `refused()` now also matches Windows' WSAECONNREFUSED rendering (`(os error 10061)`) with a doc comment explaining the three tokens — this is the exact trap #194's round-2 review independently found live on windows-latest.
- **n3/n4**: dead `exit` binding gone; `wait_bounded`/`refused` re-exported from `common` for a single import surface.

Issue #100, opened 2026-06-04
Client accepts server-only options that iperf3 rejects (IESERVERONLY) — mirror of #65


therealevanhenry:
Follow-up surfaced during #99 (#65) review. #65/#99 implemented the **client-only-on-server** direction: `-s` now rejects client-only flags like iperf3's `IECLIENTONLY`.

iperf3 also enforces the **mirror** direction: a client (`-c`) rejects **server-only** options with `IESERVERONLY` (`iperf_api.c:1817-1819`, `server_flag = 1`). riperf3 currently accepts-and-ignores them on the client.

iperf3's `server_flag` set (cross-reference against riperf3's `Cli` for the exposed subset):
- `-D/--daemon`
- `-1/--one-off`
- `--idle-timeout`
- `--server-max-duration` (`--rcv-timeout`? verify)
- `--server-bitrate-limit`
- `--rsa-private-key-path` / `--authorized-users-path` / `--time-skew-threshold` / `--use-pkcs1-padding` (server auth)
- `--pidfile`? `--logfile`? (verify against iperf3 — these may be common)

### Scope
Symmetric to #65: add `Cli::first_server_only_violation()` mirroring iperf3's `server_flag` set (for flags riperf3 exposes), reject in `main()` for the client branch before side effects, same canonical-substring error style (iperf3's `IESERVERONLY` text: verify in `iperf_error.c`). Adjudicate the exact set against `iperf_api.c` — do not guess.

Lower priority than the client-only-on-server case since server-only flags on a client are a rarer mistake, but needed for full drop-in parity.

Issue #195, opened 2026-06-11
Flake family: bidir/UDP setup starvation on loaded runners — empty-output harness timeouts (~1/20 at 2-core; post-#191 residual + uncovered bidir_intervals)


therealevanhenry:
## Residual 2-core flake: `udp_bidir_end_block_pairs_each_stream` times out (~1/20)

First seen on PR #193's `Test (Linux)` (run 27320054095): `udp_bidir_end_block_pairs_each_stream` (riperf3-cli `end_block_text`) hit the harness timeout with empty stdout/stderr ("client: timed out; stdout so far: ; stderr so far:"). Rerun passed.

**Attribution (faithful 2-core repro, `taskset -c 0,1` + `--test-threads=2`, 20 rounds each):**
- PR #193 branch: 1/20 failed (same test, same timeout shape)
- main @ a92ebd0: 1/20 failed

→ pre-existing on main, NOT introduced by #193. This is the residual of the #191 de-flake (the binary already serializes UDP tests via `udp_serial`, and this test holds the lock), so the remaining starvation is inside a single serialized UDP bidir run — most plausibly the multi-stream UDP-connect handshake or the bidir send threads starving under 2-core scheduling until the harness's deadline.

**Next probe when it recurs:** bump the harness timeout for this test and capture the client's own exit/stderr (per the ci-flake-repro discipline, the harness timeout masks the real error — a ~30 s `UDP_CONNECT_TOTAL_TIMEOUT` exit would point at the handshake, the #178/#184 class).

Repro logs: 2-core loop, failure rate ~1/20 both refs, 2026-06-11.

therealevanhenry:
New flake shape, same family (loaded-runner setup races): `tcp_bidir_end_block_pairs_each_stream` on **macos-latest** (PR #203's CI, run 27328012565) — the client died with `Connection reset by peer (os error 54)` during setup. A RST is not a refused connect, so the #176 refused-retry correctly doesn't retry it; whether the harness should also retry reset-class setup failures (bounded, pre-data) is a fair question for the next de-flake pass — a reset after data starts must stay fatal.

therealevanhenry:
## Full campaign dossier (2026-06-11, the 0.7.3 cycle — ~30 full CI rounds)

This flake hit **6 distinct CI runs and 2 local sessions** during the 0.7.3 campaign, including the version-only release bump. Compiling everything known so it stops being re-diagnosed from scratch.

### Hit log

| where | test | signature |
[...]

Issue #159, opened 2026-06-09
Windows flake: UDP bidir -J run produced zero intervals (bidir_intervals test)


therealevanhenry:
First seen on PR #150's CI re-run (windows-latest, run 27241177935): `udp_bidir_interval_sums_split_udp_stats` panicked with `client: no intervals: {...,"intervals":[],...}` — a `-u --bidir -t 2 -i 1 -J` run on the Windows runner returned a correct end block (both directions carried bytes) but an EMPTY intervals array, i.e. the interval reporter never collected a single tick over a 2-second run with a 1-second ticker.

The same test passed on every prior windows-latest run (PR #141's own CI, #148's runs, #150's first run) and on Linux/macOS/FreeBSD throughout — so this is a low-frequency Windows timing flake, not a deterministic break. Plausible mechanisms to check: tokio timer coarseness under a loaded runner vs the #55 biased end-select (end notification beating BOTH ticks?), or the Windows UDP spawn_blocking interplay starving the reporter task's first poll.

**Action:** if it recurs, instrument/harden — either make the test tolerate a degenerate zero-interval run only when wall time < interval (it shouldn't be), or chase the reporter scheduling on Windows. Filed for tracking after one rerun; not blocking #150 (unrelated PR).

therealevanhenry:
Recurrence #2 (2026-06-11): `default_rate_udp_paces_across_intervals` (udp_pacing.rs:74) on windows-latest, PR #197's CI (run 27322503216) — `expected ~3 intervals` with the intervals array empty/short on a `-u` 3s run, same shape as the original `bidir_intervals` hit. Two distinct test binaries now affected; both are short-duration UDP -J runs where the interval reporter collects nothing on a loaded Windows runner. Per the filed action, this crosses into instrument/harden territory: next Windows-touching PR (#180 in the 0.7.3 queue) should add the reporter-scheduling instrumentation or tolerate degenerate runs explicitly.

therealevanhenry:
[...]

PR #197, opened 2026-06-11
fix: --bind-dev rejects where unhonorable; server applies it like netannounce (#149)

### LINKED ISSUE #149: --bind-dev silently no-ops on FreeBSD/NetBSD; server mode never applies it
Surfaced during the #148 cold review (pre-existing on both counts; #148 changed neither).

1. **FreeBSD/NetBSD silent no-op:** `ClientBuilder::build()` rejects `bind_dev` only on `not(unix)` (client.rs ~1686), but the `set_bind_dev` fallback for non-Linux/non-macOS unix (net.rs fallback variant) silently returns `Ok(())`. A FreeBSD user passing `--bind-dev` gets no binding and no error. iperf3 compiles the option out without `CAN_BIND_TO_DEVICE` (so the flag errors as unrecognized). Faithful fix: reject `bind_dev` at `build()` on platforms where it cannot be honored, mirroring the #100/#128 rejection pattern.

2. **Server mode:** the CLI wires `--bind-dev` only in `build_client` (cli.rs ~538); the server never applies it. iperf3 binds the listening socket to the device in `netannounce`. Fix: `ServerBuilder::bind_dev` + application on the listener pre-`bind()`, matching #148's pre-connect ordering on the client side.

Both are small; could ride 0.7.2 if scope allows, otherwise backlog.

therealevanhenry:
### Silent no-op closed
`ClientBuilder::build()` rejected `bind_dev` only on `not(unix)`, while `net.rs`'s non-Linux/non-macOS unix fallback silently returned `Ok(())` — a FreeBSD/NetBSD `--bind-dev` produced no binding and no error. iperf3 without `CAN_BIND_TO_DEVICE` doesn't recognize the option at all (configure gates it on SO_BINDTODEVICE / IP_BOUND_IF), so the faithful shape is config-time rejection (the #100/#128 pattern). Both builders now reject everywhere except Linux (SO_BINDTODEVICE) and macOS (IP_BOUND_IF), and the fallback `set_bind_dev` errors instead of lying.

### Server-side application (iperf3's netannounce)
`ServerBuilder::bind_dev` + CLI wiring in `build_server`; `tcp_listen` and `udp_bind_reusable` take `bind_dev` and apply it **pre-bind** (consistent with #88's pre-connect ordering) on the TCP listener and every UDP server socket, including the recycling path's rebinds and the demux socket.

### Verification
- Linux readback tests: `SO_BINDTODEVICE` on the listener and the UDP socket (`lo`).
- Both-roles rejection test runs on the FreeBSD/NetBSD/Windows CI jobs (cfg-gated to the unsupported set).
- The existing client wiring test's gate tightened `unix` → `linux|macos` to match (it would otherwise fail on FreeBSD CI now).
- Full workspace green; clippy/fmt clean; xcheck green on all 7 targets.

Fixes #149.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ riperf3 builds on Linux, macOS, FreeBSD, and Windows. Linux is the reference pla
 | `-Z` zerocopy (sendfile) | yes | yes | yes | |
 | `-C` congestion control | yes | | yes | |
 | `-D` daemon | yes | | yes | |
-| `--bind-dev` | yes | yes | | |
+| `--bind-dev` | yes | yes (client) | | |
 | `--rcv-timeout` | yes | yes | yes | |
 | `--cntl-ka` keepalive | yes | yes | yes | |
 | TCP_INFO stats | yes | yes | yes | |
--- a/riperf3-cli/src/cli.rs
+++ b/riperf3-cli/src/cli.rs
@@ -1419,8 +1419,10 @@ mod cli_tests {
             assert_eq!(c, expected_client("h").bind_dev("eth0").build().unwrap());
         }
 
-        // #149: the server applies --bind-dev too (iperf3's netannounce).
-        #[cfg(any(target_os = "linux", target_os = "macos"))]
+        // #149: the server applies --bind-dev too (iperf3's netannounce —
+        // SO_BINDTODEVICE-only, so the server side is Linux-only; iperf3's
+        // macOS IP_BOUND_IF covers only the client path).
+        #[cfg(target_os = "linux")]
         #[test]
         fn bind_dev_wired_server() {
             let cli = Cli::parse_from(["riperf3", "-s", "--bind-dev", "eth0"]);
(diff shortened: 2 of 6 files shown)


therealevanhenry:
## Review round 1 → resolved

- **Blocker 1**: took the faithful option — `ServerBuilder` rejects `--bind-dev` everywhere but Linux (netannounce is SO_BINDTODEVICE-only; the Sept-2025 macOS IP_BOUND_IF commit covers only the client path, so a real macOS `iperf3 -s --bind-dev` fails the listen). Client keeps linux|macos. Every parity comment corrected to that ground truth.
- **Blocker 2**: moot via the rejection; the macOS native CI job now exercises the server-side rejection test (`bind_dev_rejected_on_server_off_linux`, cfg'd `not(linux)`).
- **n3**: the `-B` doc line is back on `bind_address`.
- **n4**: coverage wording fixed — FreeBSD + Windows CI jobs; NetBSD is xcheck compile-only.
- **n7**: recorded on #149 in a moment (the listener-recreation upstream quirk + `-B host%dev` shorthand).

Workspace green; clippy/fmt clean; xcheck 7/7.

therealevanhenry:
Round 2 verdict APPROVE; the two doc nits + README note applied (ClientBuilder rustdoc now states the linux|macos gate; the stale integration.rs no-op comment corrected; README macOS cell reads "yes (client)"). Awaiting the flake-rerun of windows-latest/Test(Linux) (both known classes: #159 recurrence + #195), then this is merge-ready.

PR #200, opened 2026-06-11
fix: server UDP paths apply TOS like iperf_common_sockopts (#154)

### LINKED ISSUE #154: Server UDP path never applies TOS (iperf3 applies common sockopts to UDP streams on both roles)
Noticed during #45's consistency pass: the server applies `set_tos` only to TCP data streams (server.rs TCP accept loop); the UDP path — where streams are demuxed off a shared socket — never applies TOS at all. iperf3 runs `iperf_common_sockopts` (incl. IP_TOS/IPV6_TCLASS) on UDP stream sockets on both roles, which matters for reverse/bidir where the server marks egress traffic. Fix needs thought about the demux architecture (per-stream sockets vs the shared one). Out of #45's scope (error-handling consistency); filing for the backlog.

therealevanhenry:
## #154 — server UDP paths never applied TOS

iperf3 runs `iperf_common_sockopts` (IP_TOS/IPV6_TCLASS) on UDP stream sockets on **both** roles; riperf3's server applied `set_tos` only to TCP data streams (the #45 site). Net effect: reverse/bidir UDP egress from the server was never marked.

- **Recycling path**: applied per stream socket (next to the `-w` application, pre-`into_std`), fatal per the #45 policy — identical to the TCP accept-loop site.
- **Demux path**: applied once to the shared socket; every stream shares one `cfg.tos`, so once-per-socket ≡ iperf3's once-per-stream.

### Tests (red→green)
Linux-gated egress checks drive the **real UDP connect handshake** against each private setup fn directly (no `RIPERF3_UDP_SERVER_DEMUX` env-var race) and assert the first data datagram's IP_TOS cmsg on an `IP_RECVTOS` probe client. Both observed `Some(0)` red before the fix, `Some(0x48)` after. Raw-libc cmsg walk (nix 0.29's `UnknownCmsg` fields are private); msghdr field casts are `as _` — musl's `socklen_t` vs glibc's `usize`, caught by xcheck.

Full workspace green; clippy/fmt clean; xcheck 7/7. Fleet `tcpdump -v` spot-check of `-S`-marked reverse UDP rides the post-merge drift gate.

Fixes #154.

### DIFF
--- a/riperf3/src/server.rs
+++ b/riperf3/src/server.rs
@@ -1902,7 +1902,9 @@ mod tests {
     fn recv_udp_tos(sock: &std::net::UdpSocket) -> Option<u8> {
         use std::os::fd::AsRawFd;
         let mut data = [0u8; 2048];
-        let mut cmsg = [0u8; 64];
+        // u64-aligned backing store: CMSG_FIRSTHDR/NXTHDR deref cmsghdr
+        // fields, and a bare [u8] array has alignment 1 (review r1 n6).
+        let mut cmsg = [0u64; 8];
         let mut iov = libc::iovec {
             iov_base: data.as_mut_ptr() as *mut _,
             iov_len: data.len(),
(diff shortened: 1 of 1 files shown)


therealevanhenry:
Review round 1 nits applied: PID-windowed sub-ephemeral UDP port pick replaces the bind-:0-then-drop probe (n5); the cmsg buffer is u64-aligned (n6); the IP_RECVTOS setsockopt return is asserted (n7). Tests green, clippy/fmt clean.

Issue #202, opened 2026-06-11
--timestamps FORMAT argument ignored: hardcoded HH:MM:SS UTC vs iperf3's strftime %c


therealevanhenry:
Split from #168 (its PR routes the prefix through the tee'd line path; this tracks the FORMAT fidelity):

riperf3's `--timestamps[=FORMAT]` accepts the optional format argument but renders a hardcoded `HH:MM:SS ` (UTC, no date). iperf3 passes the argument to strftime with default `"%c "` (locale date+time, local timezone).

Faithful fix: render via libc strftime (both roles route through the one prefix renderer now — `macros::output_timestamp_prefix`), defaulting to `%c `, honoring the user FORMAT. Needs a small strftime binding (nix has none; libc::strftime exists) + wiring the stored format string into the guard.

therealevanhenry:
Residue from #203's round-1 review (n4/n5), same prefix-coverage theme as the FORMAT gap: (1) `vprintln!` tees into the --get-server-output capture without the timestamp prefix (macros.rs renders only the title there) — a `-V --timestamps` capturing server returns mixed prefixed/unprefixed lines; (2) the pre-exchange banner lines ("Server listening on…", "Connecting to host…") print via bare println!/vprintln! and never get prefixed, where iperf3 timestamps every iperf_printf line. Both belong to the same strftime-prefix rework this issue tracks.

PR #203, opened 2026-06-11
fix: --get-server-output x --json-stream parity on both roles; timestamps ride the capture (#168)

### LINKED ISSUE #168: --get-server-output × --json-stream divergences (both roles) + timestamps-in-capture gap
Scoped out of #164 (per its round-1 review, confirmed live vs iperf 3.20):

1. **Server `--json-stream`**: iperf3 keeps `json_top` alive specifically for this flag (`discard_json = ... && !(role=='s' && get_server_output)`, iperf_api.c:3900) and attaches `server_output_json`. A riperf3 `--json-stream` server attaches nothing — a real iperf3 client requesting output silently gets none.
2. **Client `--json-stream`**: iperf3 emits a `server_output_text`/`server_output_json` NDJSON event before `end` (iperf_api.c:5310-5323; live-confirmed event order `start, interval, server_output_text, end`); riperf3's json-stream client drops the returned output.
3. **`--timestamps` on a capturing server**: the timestamp prefix is printed outside the tee'd line path, so the capture lacks timestamps (iperf3 buffers the prefixed `linebuffer`, so its attachment includes them) — fold the prefix into the line rendering.

therealevanhenry:
## #168 — `--get-server-output` × `--json-stream` (both roles) + timestamps-in-capture

Three live-confirmed divergences vs iperf3 3.20:

1. **`--json-stream` server attaches nothing.** iperf3 keeps `json_top` alive specifically for a requesting client (`discard_json` gates on `!(server && get_server_output)`, iperf_api.c:3900) and attaches `server_output_json`. The attach gate now includes `json_stream` — a real iperf3 client requesting output gets it.
2. **`--json-stream` client drops the returned output.** It now emits `server_output_json`/`server_output_text` NDJSON events **before** `end`, matching iperf3's event order (`start, interval*, server_output_*, end` — iperf_api.c:5310-5323).
3. **`--timestamps` on a capturing server**: the prefix printed outside the tee'd line path, so the `--get-server-output` capture lacked timestamps. The prefix now renders inside `titled()` (per **line**, like iperf3's prefixed linebuffer) via a run-scoped `OutputTimestampGuard` mirroring the title guard; the reporter's per-tick `print!` and its config field are gone. The strftime FORMAT fidelity gap is split to #202.

### Verification
TDD: three CLI tests observed red first; full workspace green; clippy/fmt clean.

Fixes #168.

### DIFF
--- a/riperf3-cli/tests/get_server_output.rs
+++ b/riperf3-cli/tests/get_server_output.rs
@@ -187,3 +187,140 @@ fn no_flag_no_server_output() {
     assert!(v.get("server_output_text").is_none());
     assert!(v.get("server_output_json").is_none());
 }
+
+// ---------------------------------------------------------------------------
+// #168: --get-server-output x --json-stream divergences (both roles) + the
+// timestamps-in-capture gap. Ground truth iperf3 3.20 (iperf_api.c:3900,
+// 5310-5323): a --json-stream SERVER keeps its JSON alive when the client
+// requests output and attaches server_output_json; a --json-stream CLIENT
+// emits server_output_text/_json NDJSON events BEFORE `end`; a capturing
+// --timestamps server's returned text carries the prefixes.
+// ---------------------------------------------------------------------------
+
+/// A `--json-stream` server attaches its JSON report for a requesting client.
+#[test]
+fn json_stream_server_attaches_json_output() {
+    let port = free_port();
+    let ps = port.to_string();
+    let server = spawn_server_capturing(&["--json-stream"], &ps);
+    let out = run_capturing(
+        &[
+            "-c",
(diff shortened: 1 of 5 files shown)


therealevanhenry:
Round 1 blockers resolved: (1) the new events route through `json_stream_event` (envelope order preserved); (2) `keep_intervals` — a json-stream server with a requesting client collects while streaming, and the test now pins non-empty attached intervals; (3) the guard is constructed only when active (`cond.then(set)`), eliminating the cross-run clobber. n4 (vprintln tee without prefix) and n5 (banner lines) folded into #202's residue notes — commenting there. Workspace green, clippy/fmt clean.

Issue #205, opened 2026-06-11
Repeated CLI flags rejected: clap errors where iperf3's getopt is last-wins


therealevanhenry:
Found while adding paced-UDP compat cells (#163): `iperf3 -u -b 0 ... -b 100M` works (getopt: the last occurrence of a flag wins), `riperf3` errors with `cannot be used multiple times`. Scripts that append override flags to a base command line — a common wrapper pattern — break against the drop-in.

Faithful fix: `#[command(args_override_self = true)]` on the clap derive (one line; applies last-wins to every arg uniformly, matching getopt). Add wired tests for a repeated `-b` and a repeated `-t`.

PR #206, opened 2026-06-11
fix: grow-only UDP SO_SNDBUF; paced-UDP compat coverage (#163)

### LINKED ISSUE #163: UDP -b at moderate rates: throughput far below target with heavy loss (fleet, -b 100M → ~16 Mbit effective, ~30% loss)

therealevanhenry:
## #163 — investigation + the residual fix

**Fleet diagnosis first** (per the issue's ask for an iperf3 same-cell comparison): the filed symptom (`-b 100M` → ~16 Mbit effective, 30-40% loss) **no longer reproduces** on main. Full sweep on the sandbox fleet, VM-to-VM:

| cell | result |
|---|---|
| iperf3 `-b 100M` baseline | 100.0 Mbit/s, 0.0% loss |
| riperf3 `-b 10M` / `-b 100M` / `-b 1G` | on target, 0.0% loss |
| riperf3 `-b 100M -w 416K` | 100.0 Mbit/s, 0.0% loss |
| riperf3 `-b 100M/1`, `-b 1G/1` (burst=1, worst sndbuf case) | on target, 0.0% loss |
| riperf3 `-b 0` (regression check) | 38.1 Gbit/s, 1.0% loss (normal) |

#190's pacing-quantum batch sizing (merged the day after #163 was filed) fixed the live mechanism.

**What lands here:**
- `configure_udp_sender` is **grow-only**: the unconditional `set_send_buffer_size` shrank SO_SNDBUF ~9-90× below the OS default whenever batch×blksize was small (quantum batch at moderate `-b`; `-b rate/1` floors it at one datagram) and clobbered an earlier `-w` bump. iperf3 never sets UDP SO_SNDBUF except for `-w`. Asymptomatic on virtio (fast TX completions) but the exact serialization hazard on completion-latency-bound paths — flagged independently by #193's round-2 review (n3). Red→green unit tests pin both properties.
- **Coverage**: the matrices only ran `-b 0` — why #163 had no signal. `scripts/interop.sh` gains `udp_b100m`/`udp_b100m_rev`; the fleet compat skill gained the same pair, so every future drift gate exercises the paced path.
- Side-find filed as #205 (clap rejects repeated flags where getopt is last-wins).

Fixes #163.

### DIFF
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -725,6 +725,7 @@ impl Client {
                         // so a low -b over a large datagram paces smoothly.
                         let pt = self.pacing_timer;
                         let bu = self.burst;
+                        let uw = self.window.is_some();
                         let u64bit = self.udp_counters_64bit;
                         let use_sendmmsg = self.sendmmsg;
                         let st = start.clone();
@@ -732,11 +733,11 @@ impl Client {
                         thread_gate.spawn(move || {
                             if use_sendmmsg {
                                 stream::run_udp_sender_sendmmsg(
-                                    std_sock, c, bs, d, rate, pt, bu, u64bit, st, md,
+                                    std_sock, c, bs, d, rate, pt, bu, uw, u64bit, st, md,
                                 )
                             } else {
                                 stream::run_udp_sender_blocking(
-                                    std_sock, c, bs, d, rate, pt, bu, u64bit, st, md,
+                                    std_sock, c, bs, d, rate, pt, bu, uw, u64bit, st, md,
                                 )
                             }
                         })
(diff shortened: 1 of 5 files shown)


therealevanhenry:
Round 1 nits applied: the senders pass `None` under an explicit `-w` (iperf3 never applies a batch-derived size; this also keeps `sndbuf_actual` truthful), the comment documents the Linux doubled-readback forgone-grow band, and a readback failure degrades to skip. Workspace green, clippy/fmt clean, xcheck 7/7.

PR #207, opened 2026-06-11
test: tolerate a dropped final interval in the UDP pacing test (#159 de-flake)

therealevanhenry:
Three windows-latest recurrences of the #159 class in one CI window (PRs #197/#200/#196), all in `default_rate_udp_paces_across_intervals`: on a loaded runner the run's catch-up burst lands after the reporter's end snapshot, so the FINAL interval is dropped — two 1-second intervals covering 0→1/1→2 with the third absent, while the end block sums the full 3 s (under-coverage evidence on #159, which stays open as the product defect). Two intervals still discriminate the #185 burst-then-starve signature (per-interval >0, <80% of total; ≥2 is the principled floor — at one interval the <80% check is unsatisfiable), so the count assertion relaxes to ≥2 rather than flaking a required check. Refs #159.

### DIFF
--- a/riperf3-cli/tests/udp_pacing.rs
+++ b/riperf3-cli/tests/udp_pacing.rs
@@ -61,12 +61,16 @@ fn default_rate_udp_paces_across_intervals() {
         .unwrap_or_else(|e| panic!("client -u -J is not valid JSON ({e}): {out}"));
     let intervals = v["intervals"].as_array().expect("intervals");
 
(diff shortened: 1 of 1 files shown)


therealevanhenry:
Round 1 APPROVE; both nits applied — the comment/commit wording now states the dropped-final-interval mechanism (with the under-coverage called out as #159's open product defect), and the truncation evidence is posted to #159.

Issue #198, opened 2026-06-11
Error-sink parity: -J error key, --logfile error line, clap exit code, no-mode exit (follow-ups to #151)


therealevanhenry:
Follow-ups from the #194 round-1 review (n8) — error-sink divergences adjacent to #151's printer, each verified against iperf3 (iperf_error.c:126-141) or live:

1. **`-J` errors**: iperf3 puts the message in the JSON `"error"` key on stdout and prints NOTHING to stderr; riperf3 with `-J` prints the stderr line and emits no JSON. (#170's ServerTerminated partial-summary work adds the `error` field plumbing this needs.)
2. **`--logfile`**: iperf3 writes the error line to the logfile when `outfile != stdout`; riperf3 always uses stderr (the dup2 redirect covers stdout only).
3. **clap usage errors** exit 2 with clap's format (`Cli::parse()` auto-exits before the #151 printer); iperf3 exits 1 with `iperf3: error - ...`.
4. **"No mode specified"** exits 0; iperf3 raises "parameter error - must either be a client (-c) or server (-s)" and exits 1.

All pre-existing; the #151 printer makes 1-2 one-liner-adjacent.

PR #208, opened 2026-06-11
fix: watch the control socket during the data phase, like iperf3's select (#170)

### LINKED ISSUE #170: Data-transfer phase gaps vs iperf3's select loop: control-death accounting, ServerTerminate summary parity, -n/-k control watch
## Context

iperf3's client main loop is a single `select()` over the control socket **and** data sockets simultaneously, so control-channel events are observed mid-transfer. riperf3's client run paths await data-stream completion and check the control socket at phase boundaries. Three behavioral gaps follow; none break the happy path or the compat matrix, but each diverges from iperf3 under failure/edge conditions:

1. **Control-connection death mid-transfer is detected late (or not at all) in some end-condition modes.** iperf3 notices the control socket closing immediately (select fires, `iperf_run_client` errors with `error - control socket has closed unexpectedly`); riperf3 can keep transmitting until the duration/byte budget is exhausted.

2. **`ServerTerminate` (server pressed Ctrl-C / `SIGTERM` mid-test) summary parity.** iperf3 receives `SERVER_TERMINATE` (state 13) on the control socket mid-transfer, stops streams, and still renders a summary from data so far (marked with the termination error). riperf3's handling at the phase boundary produces different summary/exit behavior.

3. **`-n`/`-k` (byte/block budget) runs have no concurrent control watch at all** — the sender loop runs to budget exhaustion without polling the control socket, so cases 1–2 are unobservable in those modes until the end.

## Fix sketch

A `tokio::select!` arm on the control socket (read-half watch for EOF/state bytes) racing the data-completion future in the client run paths, mapped to iperf3's behaviors: EOF → `ControlSocketClosed` error mid-test; `SERVER_TERMINATE` → stop streams, summarize partial data, exit with iperf3's message/code parity.

## Notes

- Found during PR #166 review (round 1); out of scope for the 0.7.2 review-findings batch (each item is a behavioral change needing its own TDD + cross-tool verification against real iperf3 failure injection).
- Faithfulness class: failure-mode parity with iperf3's select-driven loop.

therealevanhenry:
## #170 — control-socket watch during the data phase

iperf3's client main loop is a single `select()` over control + data fds; riperf3 checked the control socket only at phase boundaries, and only in duration mode. Three failure-mode gaps closed:

1. **Control death mid-test** → prompt `ControlSocketClosed` (IECTRLCLOSE text) in **every** end-condition mode. Duration mode previously swallowed the error and "completed" at the full `-t` (failing late as broken-pipe Io); `-n`/`-k` had no watch and hung forever. The watch also fixes the latent "any stray state byte ends the wait" bug — non-terminate states log and continue, like iperf3's no-op handling.
2. **SERVER_TERMINATE mid-test** → partial summary from local data (iperf3 flips to DISPLAY_RESULTS), then `ServerTerminated` (IESERVERTERM). `-J` carries the blob `"error"` key per `iperf_json_finish`; `--json-stream` emits the `error` event before `end` (via the existing envelope helper). The stringly `Aborted("server terminated")` migrates to the typed variant (additive, `#[non_exhaustive]`).
3. Outcomes surface only **after** the #147 cleanup — the reporter-leak class can't regrow (pinned by the updated Arc-refcount test).

### Verification
TDD: 3 new mock-server tests red first (wrong variant; the `-n` hang at its 8s timeout; missing summary) → green. The CLI prints `riperf3: error - the server has terminated` / `control socket has closed unexpectedly` via #151's printer. 300 lib tests; full workspace green; clippy/fmt clean.

Fixes #170.

### DIFF
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -176,6 +176,10 @@ async fn watch_control(ctrl: &mut tokio::net::TcpStream) -> ControlEvent {
             Ok(other) => {
                 log::debug!("ignoring control state {other:?} during the data phase");
             }
+            // Recorded deviation (r1 n3): iperf3 splits EOF (IECTRLCLOSE) /
+            // read error (IERECVMESSAGE) / unknown state byte (IEMESSAGE);
+            // riperf3 folds all three into the closed class — the headline
+            // kill case (FIN→EOF) matches byte-for-byte.
             Err(_) => return ControlEvent::Closed,
         }
     }
@@ -980,6 +984,18 @@ impl Client {
         // -n/-k modes had no watch at all, control death in duration mode
         // "completed" the test, and any stray state byte truncated the wait.
         let mut control_event: Option<ControlEvent> = None;
+        // Watch-arm end time (#170 review r1 n2): the reporter timeline is
+        // post-omit-rebased (iperf3 restamps start_time at the boundary), so
+        // an event past the warm-up reports `elapsed - omit`; during the
+        // warm-up the raw elapsed matches iperf3's un-restamped clock.
+        let omit_secs = self.omit as f64;
+        let watch_end_secs = move |raw: f64| {
+            if raw > omit_secs {
+                raw - omit_secs
+            } else {
+                raw
+            }
+        };
         let end_secs = match end_condition {
             EndCondition::Duration(dur) => {
                 // The UDP senders also enforce this deadline themselves inside
@@ -997,7 +1013,7 @@ impl Client {
                     _ = tokio::time::sleep(wall) => dur.as_secs_f64(),
                     ev = watch_control(ctrl) => {
                         control_event = Some(ev);
-                        report_start.elapsed().as_secs_f64()
+                        watch_end_secs(report_start.elapsed().as_secs_f64())
                     }
                 }
             }
@@ -1006,7 +1022,7 @@ impl Client {
                     secs = self.wait_byte_limit(streams, target, &report_start, omit_boundary.as_ref()) => secs,
                     ev = watch_control(ctrl) => {
                         control_event = Some(ev);
-                        report_start.elapsed().as_secs_f64()
+                        watch_end_secs(report_start.elapsed().as_secs_f64())
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


therealevanhenry:
## Review round 1 → resolved

- **Blocker 1**: all three fabrication sites fixed (JSON cross-grafts gated off under error; UDP terminated sender keeps its local packet count; the text pair synthesizes a ZEROED opposite-role half — T2 now pins the receiver line).
- **Blocker 2**: the watch arms rebase past the omit boundary (`elapsed - omit`; raw during warm-up, matching iperf3's un-restamped clock).
- **n3**: the IECTRLCLOSE/IERECVMESSAGE/IEMESSAGE fold is a recorded deviation in `watch_control`. **n4**: the `-J` stderr double-report is exactly #198 item 1 (errors-into-blob) — noted there. **n5**: dup allow removed. **n6**: pre-existing arm, noted.

therealevanhenry:
## Review round 2 → resolved (9bfc191)

Both remaining fabrication axes fixed with the same error-gate pattern: the bidir `-J` sum cross-grafts go to 0 on a terminated run (matching your live iperf3 capture), and a terminated reverse-UDP receiver stream reports `bytes: 0` while keeping its measured packets — with `sum.packets` now carrying iperf3's receiver-count fallback (iperf_api.c:4242), which is reachable exactly when a terminated `-R` run never exchanged. Workspace green, clippy/fmt clean.

therealevanhenry:
Round 3's blocking item resolved: three deterministic unit pins for the terminated-run shapes (`terminated_bidir_sums_zero_the_absent_peer_halves`, `terminated_reverse_udp_keeps_measured_packets_with_zero_bytes`, `peer_absent_without_error_still_grafts`), exercising `error: Some` in json_report's harness for the first time. The branch is also updated onto main (the #168 union puts the NDJSON tail in iperf3's order: error → server_output_* → end). The three pre-existing notes (UDP-bidir tcp_sum aggregates, json-path end-sum jitter, reverse-UDP per-stream bytes) get filed as backlog issues. Workspace green, clippy/fmt clean.

Issue #158, opened 2026-06-09
Signal-exit follow-ups: second-signal handling, Windows pidfile/Ctrl+C, panic-path unlink


therealevanhenry:
Deferred items from the #150 reviews (the merged behavior is strictly better than pre-#105 and verified iperf3-parity on the unix paths):

1. **Second signal during shutdown is silently swallowed** — after the first SIGTERM resolves the select, further signals during the (brief) teardown window do nothing; pre-#150 a second Ctrl-C always killed the process. Bounded (teardown is sub-ms in practice; up to ~10s with a still-blasting UDP peer via UDP_RECEIVER_DRAIN_TIMEOUT). Most daemons treat a second signal as immediate-exit; consider that. Also unrecorded: the sigend path doesn't dump mid-test stats or send SERVER_TERMINATE like iperf3's iperf_got_sigend (scope-noted on the PR; belongs here for tracking, including the -J output-shape difference: iperf3 emits the JSON document with an "error" field on sigend).
2. **Windows: pidfile stranded on Ctrl+C** — the pidfile is written on all platforms but the signal path is unix-only; iperf3 installs SIGINT/SIGTERM via CRT signal() on Windows too. Needs tokio::signal::windows wiring + the same unlink ordering.
3. **Panic inside block_on bypasses the unlink** (panic=unwind) — a RAII drop-guard on the pidfile path would cover it for free. iperf3 leaks on crash too, so parity-neutral; nice-to-have.

therealevanhenry:
0.7.3 status (PR #211, merged ded769a→707cb87): items 1-3 are DONE — Windows signal wiring (Ctrl+C/Break/close/logoff/shutdown via tokio ≥1.44's parked HandlerRoutine), the pidfile RAII drop-guard (panic-path unlink), and second-signal immediate exit (async-signal-safe handler, dies by the signal, test-pinned). The one remaining bullet — the sigend mid-test stats dump + SERVER_TERMINATE send — is tracked as #210 (0.7.4). This issue can close in favor of #210.

therealevanhenry:
Closing per the status comment above: items 1-3 shipped in 0.7.3 (PR #211); the sigend stats-dump/SERVER_TERMINATE remainder continues as #210 (0.7.4).

PR #211, opened 2026-06-11
fix: Windows signal wiring, pidfile RAII guard, second-signal immediate exit (#158)

therealevanhenry:
## #158 — signal-exit follow-ups (items 1-3)

- **Windows signal wiring**: `Sigend` analog over `tokio::signal::windows` (Ctrl+C/Break/close/shutdown — iperf3 installs via CRT `signal()` there too), same pidfile-unlink ordering; the main select widens `unix` → `any(unix, windows)`.
- **Pidfile RAII guard**: panic=unwind can no longer strand the file (iperf3 leaks on crash; strictly better, parity-neutral).
- **Second signal during teardown exits immediately**: unix re-registers raw libc handlers (async-signal-safe body: `write`+`unlink`+`_exit`) — a runtime-spawned watcher would be cancelled at the start of the exact rt-drop hang it exists to escape (a still-blasting UDP peer holds the shared drain up to 10 s). Windows keeps a best-effort runtime watcher; close/shutdown events carry the OS's ~5 s force-kill regardless.
- The sigend **stats-dump + SERVER_TERMINATE** parity item is split to its own issue (needs a shutdown channel into the run loops; the #170 partial-summary plumbing is its natural landing).

### Verification
xcheck 7/7 (the Windows arm compiles); pidfile/daemon CLI tests green; full workspace green; clippy/fmt clean. Windows runtime check on mithrandir rides the review round.

Refs #158 (stats-dump remainder tracked separately).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -448,9 +448,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.164"
+version = "0.2.168"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "433bfe06b8c75da9b2e3fbea6e5329ff87748f0b144ef75306e674c3f6f7c13f"
+checksum = "5aaeb2981e0606ca11d79718f8bb01164f1d6ed75080182d3abf017e6d244b6d"
 
 [[package]]
 name = "libm"
@@ -1110,9 +1110,9 @@ dependencies = [
 
 [[package]]
 name = "tokio"
-version = "1.41.1"
+version = "1.44.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "22cfb5bee7a6a52939ca9224d6ac897bb669134078daa8735560897f69de4d33"
+checksum = "e6b88822cbe49de4185e3a4cbf8321dd487cf5fe0c5c65695fef6346371e9c48"
 dependencies = [
  "backtrace",
  "bytes",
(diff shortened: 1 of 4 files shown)


therealevanhenry:
## Review round 1 → resolved

- **Blocker 1**: tokio 1.41.1 → 1.44.2 (the handler-park fix, tokio #7122) — the Windows Sigend doc now describes the real semantics. libc rides at 1.44's exact floor (0.2.168): 0.2.186 breaks nix-0.29/mio-1.0.2 on NetBSD (xcheck caught it).
- **Blocker 2**: `second_signal_during_teardown_exits_immediately` lands your validated recipe as a race-guarded CLI test (bounded exit + pidfile gone hold on either race outcome).
- **n3-n7**: ctrl_logoff added; `drop(sigend)` comment reworded; SIG_DFL+raise (die by the signal); Release/Acquire on PIDFILE; cfg suppressor corrected; the Windows watcher comment credits the fall-through (n9).

xcheck 7/7, workspace green, clippy/fmt clean.

therealevanhenry:
## Review round 2 → resolved

- **F1**: the test now asserts death-by-SIGTERM (`status.signal() == Some(SIGTERM)`) — your revert-validated discriminator; the race-guard arm and its comment are gone (n4).
- **F2**: `pthread_sigmask(SIG_UNBLOCK)` after the raise — the pending-signal trap explained in the SAFETY comment; `_exit(1)` is now genuinely a fallback.
- **F3**: `tokio = "1.44"` floored in the workspace manifest with the libc hold-back note (n5).

therealevanhenry:
## Windows CI red → root-caused and fixed (not the tokio bump)

Reproduced on mithrandir: `thread 'main' has overflowed its stack` after the server banner. This PR made Windows run `async_main` inside the main `tokio::select!` for the first time — select polls futures **in place**, so the whole server state machine landed inline on Windows' 1 MiB main-thread stack (unix's 8 MiB masked it). `Box::pin` fixes it; bisect on the VM showed the overflow persisted on tokio 1.41, so the 1.44 floor and the full event set stay (r2 F3 intact). Native verification: bidir_intervals 5/5 + the UDP suite green on mithrandir with 1.44.2.

PR #212, opened 2026-06-11
fix: last-wins repeated flags like getopt; omit boundary keeps TCP_INFO extremes (#205, #199)

therealevanhenry:
### #205 — repeated flags are last-wins
`args_override_self = true` on the clap derive: iperf3's getopt takes the last occurrence (`-b 0 … -b 100M` works), and wrapper scripts appending override flags broke against the drop-in. Wired tests pin `-b` and `-t`.

### #199 — `-O` boundary keeps the TCP_INFO extremes
`iperf_reset_stats` (iperf_api.c:3669-3695) clears byte/packet counters, jitter, and the retransmit baseline — but `stream_max_snd_cwnd`/`snd_wnd`/`rtt` (and even the RTT mean's running sum) keep their warm-up peaks. riperf3's boundary fully reset the extremes accumulator, under-reading every `max_*` after an omitted warm-up. The fix is a deletion matching the cited C; a direct red-test needs a TCP_INFO injection seam the reporter doesn't have (recorded for the post-0.8.0 refactor — the change is a deletion whose behavior the reviewer can verify against the C).

Fixes #205. Fixes #199.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #215, opened 2026-06-11
fix: render --timestamps with strftime and the user FORMAT (#202)

### LINKED ISSUE #202: --timestamps FORMAT argument ignored: hardcoded HH:MM:SS UTC vs iperf3's strftime %c

therealevanhenry:
## #202 — `--timestamps[=FORMAT]` fidelity

The renderer ignored the FORMAT and printed hardcoded `HH:MM:SS` UTC; iperf3's `iperf_printf` renders **localtime through strftime** with the stored format, default `"%c "` (trailing space inside the format; user formats verbatim — iperf.h's `TIMESTAMP_FORMAT`).

- The run-scoped guard (from #203) now stores the format; unix renders via `localtime_r` + `strftime`.
- Windows keeps the HH:MM:SS fallback — the libc crate's msvc surface has no strftime (verified against 0.2.164 and latest), and native Windows has no iperf3 ground truth to match (documented in-code).
- The CLI's bare `--timestamps` already carried `default_missing_value = "%c "`; it now takes effect. Composes with #212's bare-repeat reset pin.
- Tests: a macros unit pin (format honored, cleared on drop, `%c` non-empty) + the capture CLI test moves to an explicit format (locale `%c` has no stable shape to assert).

The remaining prefix-coverage residue — banner lines and `vprintln!` output never carry the prefix, where iperf3 prefixes every `iperf_printf` line — is split to **#216** (review r1 finding 1).

Fixes #202 (strftime/FORMAT rendering; the line-coverage residue continues in #216).

### DIFF
--- a/riperf3-cli/tests/get_server_output.rs
+++ b/riperf3-cli/tests/get_server_output.rs
@@ -288,7 +288,9 @@ fn json_stream_client_emits_server_output_event_before_end() {
 fn timestamped_server_capture_carries_prefixes() {
     let port = free_port();
     let ps = port.to_string();
-    let server = spawn_server_capturing(&["--timestamps"], &ps);
+    // Explicit format: the faithful default is "%c " (locale datetime, e.g.
+    // "Wed Jun 11 ..."), which has no stable shape to assert (#202).
+    let server = spawn_server_capturing(&["--timestamps=%H:%M:%S "], &ps);
     let out = run_capturing(
         &[
             "-c",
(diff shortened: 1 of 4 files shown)


therealevanhenry:
Round 2's blocker fixed: `then_some` replaces the side-effect-free closure at both call sites (clippy's own suggestion, gating identical), and the `set()` doc now describes the format-era API while keeping the construct-only-when-active invariant text. Clippy clean with the exact CI invocation; workspace green.

PR #201, opened 2026-06-11
fix: bidir text interval rows — role tags, per-direction SUMs, sender packet counts (#143, #187)

therealevanhenry:
## #143 + #187 — bidir text interval rows match iperf_print_intermediate

Audited the C first: iperf3 prints intervals per **direction** (`current_mode` pass) — every row carries the bidir role tag (`mbuf`), each direction prints **its own** `[SUM]` (tagged), the text SUM is gated on `num_streams > 1` per direction, and a cross-direction aggregate never exists. riperf3 printed untagged rows plus one combined `[SUM]` mixing both directions, even at bidir P=1.

- Role tags on interval rows and SUMs via the #189 `fmt_id_role` plumbing (`[ 5][TX-C]` shape — identical to iperf3's `[%3d]%s`); reporter config gains `bidir`/`is_server`.
- Per-direction SUMs gated on the direction's stream count: bidir P1 → no SUM (was: mixed SUM), bidir P2 → `[SUM][TX-C]` + `[SUM][RX-C]`, non-bidir unchanged. Sending direction carries Retr (TCP) / sent count (UDP); receiving direction carries loss + mean jitter (#142 semantics).
- **UDP sender interval rows** now print iperf3's `report_bw_udp_sender_format` shape (blank jitter/loss region + trailing sent-datagram count) — the #187 ground-truth capture shows riperf3 was printing bare TCP-style rows for senders.

### Verification
- TDD: both new CLI tests observed red (missing tags + mixed SUM at P1; no per-direction SUMs at P2) → green.
- Full workspace green; clippy/fmt clean; xcheck 7/7. Live text-capture diff vs iperf3 3.20 on the fleet rides the post-merge drift gate.

Fixes #143. Fixes #187.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #217, opened 2026-06-11
fix: per-tick text separator at -P > 1; --json-stream-full-output (#204, #213)

### LINKED ISSUE #204: Text intervals: missing per-tick '- - - - -' separator at -P > 1 (iperf3 report_bw_separator)
From PR #201's round-1 review (n8), pre-existing: iperf3 prints the `- - - - -` separator line BETWEEN interval ticks when `num_streams > 1` (print_interval_results: first stream of a non-first tick prints `report_bw_separator`). riperf3 never prints per-tick separators — only the end-block one.

Faithful fix: in the interval reporter, before emitting a non-first tick's rows when the -P count > 1, print the separator. Note the #201 tests anchor the interval section on `split("- - - - -")` — fixing this requires re-anchoring those on the end-block header instead (flagged in the same review, n7).

### LINKED ISSUE #213: --json-stream-full-output unimplemented (third leg of discard_json)
From #203's round-2 review (n3): iperf3's `--json-stream-full-output` opts back into the monolithic -J blob alongside the NDJSON stream (`json_stream_full_output` is the second leg of the discard_json condition, iperf_api.c:3900); riperf3's clap rejects the flag. Small: a passthrough flag + keep_intervals + print the full blob after the stream (iperf_api.c:5325's print_full_json).

therealevanhenry:
### #204 — per-tick `- - - -` separator
`iperf_print_intermediate` (first stream of a tick): first interval → header; else if `num_streams > 1` → `report_bw_separator` (iperf_api.c:4695-4701). The gate maps to the per-direction counts that #201's SUM logic established — bidir P=1 prints two rows per tick but `num_streams==1`, so no separator, matching iperf3. **Live A/B** on identical `-t 2 -O 1 -i 1 -P 2` runs: 3 separators in both tools; iperf3 does **not** reprint the header after the omit boundary (probed live — the restamp does not retrigger the first-interval comparison), and neither do we.

### #213 — `--json-stream-full-output`
The third leg of discard_json (iperf_api.c:3900): intervals retained on both roles, and the complete monolithic document prints after the `end` event (iperf_api.c:5323's `print_full_json`). Flag without `--json-stream` is a getopt-faithful no-op.

Red-first tests: separator shape at P=2 / P=1-negative; trailing-document presence with populated intervals / absence-negative; builder wired test.

Fixes #204. Fixes #213.

### DIFF
--- a/riperf3-cli/src/cli.rs
+++ b/riperf3-cli/src/cli.rs
@@ -274,6 +274,11 @@ pub struct Cli {
     #[arg(long)]
     pub json_stream: bool,
 
+    /// With --json-stream, also print the complete monolithic JSON document
+    /// after the stream ends
+    #[arg(long = "json-stream-full-output")]
+    pub json_stream_full_output: bool,
+
     /// Enable UDP GSO/GRO
     #[arg(long)]
     pub gsro: bool,
@@ -557,6 +562,9 @@ impl Cli {
         if self.json_stream {
             builder = builder.json_stream(true);
         }
+        if self.json_stream_full_output {
+            builder = builder.json_stream_full_output(true);
+        }
         if self.udp_counters_64bit {
             builder = builder.udp_counters_64bit(true);
         }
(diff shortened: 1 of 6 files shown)


therealevanhenry:
Round 2's blocker fixed: the assertion is count-tied — T interval `[SUM]` rows before the end-block separator ⇒ exactly T separators (T-1 per-tick + the end-block one), with a ≥2-ticks floor preserving the discriminating power. 5× green under `taskset -c 0,1 --test-threads=2`.

PR #219, opened 2026-06-11
fix: --timestamps prefix covers vprintln and the server banner (#216)

### LINKED ISSUE #216: --timestamps prefix misses vprintln and banner lines (iperf_printf prefixes every line)
Residue split from #202 (whose strftime/FORMAT rendering lands in #215): iperf3 timestamps **every** `iperf_printf` line — including `Connecting to host …` (-V), the `-----` separators, and `Server listening on …` (iperf_api.c:995/1017, iperf_server_api.c:137). riperf3's prefix renders only inside `titled()` report lines; `vprintln!` output (macros.rs ~187-200) and the server banners (server.rs ~214/236, printed pre-guard) never carry it. Verified live in #215's review: `-V --timestamps=%T-` prefixes every reporter line but not the connect banner. Fix shape: route the banner/verbose prints through the prefixed path and set the guard before the server banner prints.

therealevanhenry:
## #216 — prefix line coverage

iperf3 prefixes **every** `iperf_printf` line (banner at iperf_server_api.c:137, verbose lines, separators); riperf3 rendered the prefix only inside `titled()` report lines (#215's review found this live: `-V --timestamps=%T-` left `Connecting to host …` bare).

- `vprintln!` prepends the timestamp prefix before the title prefix (iperf_printf's order).
- The server banner routes through a prefixed `banner_line`; the timestamp guard moves to `run()` scope **before** the banner (the per-test guard structurally couldn't cover it). `--timestamps` is the server's own un-exchanged flag, so run scope is its natural home.
- The banner stays out of the `--get-server-output` capture — iperf3's banner prints before the test's capture exists.

Red-first with a literal strftime format (`TSTAMP ` — no `%` directives, renders verbatim) for deterministic assertions.

Fixes #216.

### DIFF
--- a/riperf3-cli/tests/get_server_output.rs
+++ b/riperf3-cli/tests/get_server_output.rs
@@ -335,3 +335,72 @@ fn regex_lite_timestamp(line: &str) -> bool {
         && b[4].is_ascii_digit()
         && b[5] == b':'
 }
+
+/// #216: iperf3 prefixes EVERY iperf_printf line — the server's listening
+/// banner and verbose lines included (iperf_api.c:995/1017,
+/// iperf_server_api.c:137). A literal strftime format (no % directives)
+/// renders verbatim, making the assertion deterministic.
+#[test]
+fn server_banner_carries_the_timestamp_prefix() {
+    let ps = free_port().to_string();
+    let server = spawn_server_capturing(&["--timestamps=TSTAMP "], &ps);
+    let _ = run_capturing(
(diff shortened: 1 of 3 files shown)


therealevanhenry:
Round 1 APPROVE — thanks for the byte-level capture work (the \\n-carries-the-prefix quirk match is a great catch to have pinned in the record). The missing-line classes from n2 are filed as #222; the n1 spawn-helper polish rides the next commit if a round-2 nit pass happens, else as-is.

PR #223, opened 2026-06-11
feat: iperf_got_sigend parity — mid-test stats dump + TERMINATE send on both roles (#210)

### LINKED ISSUE #210: Sigend parity: dump mid-test stats and send SERVER_TERMINATE on signal, like iperf_got_sigend
Split from #158 (its PR lands the Windows wiring, pidfile RAII, and second-signal exit): iperf3's `iperf_got_sigend` mid-test (1) dumps the stats gathered so far (its -J emits the document with an `error` field — the #170 plumbing now exists for that shape) and (2) sends SERVER_TERMINATE/CLIENT_TERMINATE on the control socket so the peer ends cleanly. riperf3's sigend path just tears down. Needs a shutdown signal into `Client::run`/`Server::run` (a watch channel the select arms race, mapped to the same partial-summary path #170 built). Cross-cutting; deliberately not rushed into 0.7.3's #158 PR.

therealevanhenry:
## #210 — the #158 remainder

iperf_got_sigend (iperf_api.c:5144-5188), all three legs:

1. **Dump**: a signal mid-test renders the accumulated stats through the #170 partial-summary path on both roles (client: sender + zeroed receiver; server: its local half, measured-elapsed window — live iperf3 stamps `0.00-2.00` for a `-t 10` run killed at 2 s).
2. **Terminate send**: client → CLIENT_TERMINATE, server → SERVER_TERMINATE on the control socket. **Cross-tool live**: riperf3's CLIENT_TERMINATE makes a real iperf3 server print `iperf3: the client has terminated`; an iperf3 client's terminate makes riperf3's server dump + print the same sentence (no `error -` prefix, matching iperf3's IECLIENTTERM rendering).
3. **Exit-normal**: the CLI prints `riperf3: interrupt - the <role> has terminated by signal …` and exits 0 on TERM/INT/HUP.

### Architecture
- Additive lib API: `ClientBuilder::interrupt` / `ServerBuilder::interrupt` + chainable `with_interrupt` (a `watch::Receiver<Option<String>>` carrying the caller-formatted message — it feeds the `-J` `error` key like iperf_exit). Newtyped for the PartialEq glue-test convention.
- The server's CLIENT_TERMINATE arm previously early-returned **without the dump and leaking the reporter** (the #147 class) — now it breaks into the normal cleanup + dump, skipping only the results exchange.
- The CLI's first signal fires the watch and bounded-awaits the dump (5 s); the **hard second-signal handler arms before the dump window** (previously only covered rt-drop). The #158 test's premise changed with the graceful path — it now wedges the server with a half-open control connection (no interrupt-aware await in the param exchange), keeping the death-by-SIGTERM discriminator deterministic.

Red-first CLI tests (both directions); 3× stable + 2-core pin; xcheck 7/7; workspace green.

Fixes #210. Refs #158, #170.

### DIFF
--- a/riperf3-cli/src/main.rs
+++ b/riperf3-cli/src/main.rs
@@ -141,20 +141,57 @@ fn run() -> std::result::Result<(), Box<dyn std::error::Error>> {
     let _pidfile_guard = PidfileGuard(cli.pidfile.clone().map(std::path::PathBuf::from));
     let pidfile = cli.pidfile.clone();
     let is_server = cli.server;
+    // #210: the first signal no longer cancels the run outright — it fires
+    // this watch with iperf3's formatted message, and the lib dumps the
+    // accumulated stats + tells the peer (CLIENT_TERMINATE /
+    // SERVER_TERMINATE) like iperf_got_sigend, then returns. The bounded
+    // wait below keeps a wedged teardown from hanging the exit; the
+    // second-signal hard path remains underneath.
+    let (interrupt_tx, interrupt_rx) = tokio::sync::watch::channel::<Option<String>>(None);
     #[cfg(any(unix, windows))]
     let outcome = rt.block_on(async {
         // Box::pin: select! polls its futures IN PLACE, and async_main's
         // future is the entire client/server state machine — inline it
         // overflows Windows' 1 MiB main-thread stack (unix's 8 MiB masked
         // it). Heap-pin the big one; the signal future is tiny.
-        let mut app = Box::pin(async_main(cli));
+        let mut app = Box::pin(async_main(cli, interrupt_rx));
         tokio::select! {
             r = &mut app => r,
-            sig = sigend.recv() => Ok(Exit::Signal(sig)),
+            sig = sigend.recv() => {
+                // The SECOND-signal hard exit (#158) must cover the dump
+                // window below, not just rt-drop — register the raw handler
+                // BEFORE awaiting the dump (the libc overwrite wins over
+                // tokio's still-registered sigaction). Windows: the
+                // best-effort watcher, spawned for the same window.
+                #[cfg(unix)]
+                second_signal_exits_immediately(pidfile.as_deref());
+                #[cfg(windows)]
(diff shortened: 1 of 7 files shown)


therealevanhenry:
## Review round 1 → resolved

- **f1**: `build_report_input` gains the error param; one `report_error` feeds the `-J` doc, the json-stream end path, and the get-server-output prebuilt. Pinned by `server_json_doc_carries_the_terminate_error_key` (doc `error` = IECLIENTTERM, stderr empty). The ClientTerminated stderr line and the CLI interrupt notice gate off in JSON modes (**f3**).
- **n2**: SERVER_TERMINATE is handled in any state (the ExchangeResults-wait race dumps + IESERVERTERM, like iperf_handle_message_client).
- **n7**: the stale literal fixed. **n8**: the server_terminated full-exchange path + dead server_results get a backlog issue.
- **n4** (the 5 s idle burn): accepted-by-design this round; wiring `wait_interrupt` into the connect/idle awaits noted as a follow-up.

xcheck 7/7, workspace green, clippy/fmt clean.

therealevanhenry:
Round 2's blocker fixed (one commit): the server's json-stream branch emits the discrete `error` event before `end` like iperf_json_finish — mirroring the client's branch — pinned by `server_json_stream_emits_the_error_event_before_end` (order + stderr silence). The non-blocking client `-J` stderr leak is filed as #225. Workspace green, clippy/fmt clean.

therealevanhenry:
Post-r3 CI fixes (both verified locally): the SemVer break — `watch::Receiver` is not `UnwindSafe`, stripping the marker from all four pub types — is fixed with manual marker impls on `InterruptWatch` (panic-consistent: poll-only usage, version-counter + Arc'd slot state; `cargo semver-checks` now passes); and the #204 tick-floor assert (merged via #217, flaked here on a starved windows runner printing ONE tick from a `-t 3` run) softens to a SKIP below 2 printed ticks — the count-tied rule still pins on every healthy run.

PR #244, opened 2026-06-11
fix: reporter end-race — the final flush runs after the senders stop (#159)

### LINKED ISSUE #159: Windows flake: UDP bidir -J run produced zero intervals (bidir_intervals test)

therealevanhenry:
## #159: the reporter end-race — flush after the senders stop

The issue's root-cause comment called for exactly this: *"the faithful fix (intervals always cover the run, like iperf3's synchronous end reporting) needs the flush to happen after the senders are stopped/joined, which reorders the #55 authoritative-end-time design — a deliberate redesign, not a ride-along."*

**Old order:** `finish(end_secs)` → final flush (gated `residual_bytes > 0` at the snapshot) → `done` → senders stop. A starved sender's catch-up burst landing between the snapshot and `done` was invisible to the intervals but counted by the END block — the array under-covered the run it accounts (the windows-latest dropped-tick family, #207's forensics). With *everything* starved into the catch-up, the residual gate skipped the flush outright — the original empty-intervals shape.

**New order** (both roles): `done` → the 100 ms teardown grace, moved up (the catch-up lands in the counters) → `finish` → flush. iperf3 reads its counters after `pthread_join`; same semantics. The `[last_boundary, end_secs]` **window stays authoritative** (#55) — late-landing bytes belong to the window they were sent in, and the full suite (including the #55 pins) is green.

The reporter's tick arm previously exited bare on observing `done` (correct under the old order, a dropped final interval under the new): it now waits (bounded, 2 s) for the `finish` signal and re-arms the notify so the end arm — which owns the flush — runs. An error path that never calls `finish` still tears down promptly with no invented interval (the pre-existing semantic, pinned).

### TDD

Component-level, deterministic, in-crate (the reporter is `pub(crate)`):
- `tick_observing_done_waits_for_finish_and_flushes` — **red on main** (1000 collected vs 1500 sent)
- `catchup_bytes_after_done_land_in_the_final_interval` — pins the new driver-order contract
- `bare_done_without_finish_exits_without_inventing_intervals` — the error-path negative

### Verification beyond the suite

- 20× 2-core rounds (`taskset -c 0,1`, `--test-threads=2`) on the udp_pacing + bidir_intervals binaries: posted below.
- Cloud hammer on this branch (it inherits hammer.yml), **windows-latest** — where this family lived: posted below.

Closes #159.

### DIFF
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -1276,8 +1276,15 @@ impl Client {
         // the final interval or the summary (#55). The reporter prioritises this
         // `finish` over `done` (see its select), so the final interval still
         // flushes; we then wait for it before tearing the streams down.
-        reporter_end.finish(end_secs);
+        // #159: stop the senders FIRST and give their in-flight catch-up the
+        // teardown grace to land in the counters, THEN signal the flush —
+        // iperf3 reads its counters after the threads join, so the intervals
+        // always cover what the END block accounts. The [last_boundary,
+        // end_secs] window stays authoritative (#55) — late-landing bytes
+        // belong to the window they were sent in.
         done.store(true, Ordering::Relaxed);
+        tokio::time::sleep(Duration::from_millis(100)).await;
+        reporter_end.finish(end_secs);
         if let Some(handle) = interval_handle {
             let _ = handle.await;
         }
--- a/riperf3/src/client.rs
+++ b/riperf3/src/client.rs
@@ -1271,11 +1271,9 @@ impl Client {
             }
         };
 
-        // End of test: hand the reporter the authoritative end time, then stop
-        // the senders immediately (`done`) so no bytes leak past the deadline into
-        // the final interval or the summary (#55). The reporter prioritises this
-        // `finish` over `done` (see its select), so the final interval still
-        // flushes; we then wait for it before tearing the streams down.
+        // End of test (#55 window, #159 order): stop the senders, let the
+        // catch-up land, then hand the reporter the authoritative end time —
+        // the flush below reads settled counters.
         // #159: stop the senders FIRST and give their in-flight catch-up the
         // teardown grace to land in the counters, THEN signal the flush —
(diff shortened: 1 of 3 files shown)


therealevanhenry:
Verification record: local 2-core rounds (`taskset -c 0,1`, `--test-threads=2`) — **40/40 green** (20 rounds × udp_pacing + bidir_intervals). Cloud hammer on this branch, **windows-latest** (where the #159 family lived): [run 27374192203](https://github.com/therealevanhenry/riperf3/actions/runs/27374192203) — 8 shards × 3 iterations across udp_pacing/bidir_intervals/end_block_text, **24/24 green**. The r1 reviewer's live probe also found the post-PR property (intervals total == END exactly) is *stricter than pinned iperf3 3.21's own behavior* — GT's intervals under-covered its END by 21% in the same geometry.

Issue #345, opened 2026-07-06
post-cookie send_state failure takes a raw Io message where GT emits IESENDMESSAGE


therealevanhenry:
## Divergence

If the server's post-cookie `send_state(ParamExchange)` write fails (an RST landing between the cookie read and the state write), the error rides the serve loop's generic arm with a raw Io message. GT emits IESENDMESSAGE's sentence: "unable to send control message - port may not be available, the other side may have stopped running, etc." (iperf_error.c:305-308, set via iperf_accept's `iperf_set_send_state` failure path).

Rare race; the sink SHAPE (skeleton -J doc / one stderr line, exit 0, keep serving) is already correct since the #330-remainder PR — only the message class diverges.

## Fix shape

Map control-write failures in the pre-test phase to an IESENDMESSAGE-class error with GT's sentence.

therealevanhenry:
GT 3.21 live-probed (probe345.py, cookie + SO_LINGER(0) RST before the state write; N=12 per mode):

**Text**: deterministic 12/12 — stderr `iperf3: error - unable to send control message - port may not be available, the other side may have stopped running, etc.: Transport endpoint is not connected` (IESENDMESSAGE sentence + LIVE errno strerror, ENOTCONN — a real deterministic errno, not the stale class), rc=0, keep-serving semantics (one-off exits 0).

**-J**: silent stderr + SKELETON doc (connected:[]/version/system_info + intervals:[] + end:{}) with error key `error - unable to send control message - ...: Transport endpoint is not connected`, rc=0. NOTE: 4/12 runs the RST lost the race — the state write succeeded and the run took the params-read class instead (sink-bypassing `warning: Failed to read JSON data size - read returned -2; errno=104` + the #339 IERECVPARAMS surface). Both classes must stay reachable; the fix maps the SEND failure only.

riperf3 fix shape: catch the send_state(ParamExchange) failure in negotiate_test → new IESENDMESSAGE-class error rendering GT's sentence; the errno suffix rides the io::Error Display (#151 convention: trailing "(os error N)" divergence accepted). Wire-back: GT relays SERVER_ERROR + IESENDMESSAGE via cleanup_server on this path — include the relay (best-effort; usually unobservable, the peer RST). Same emit_pretest_error sink as #339 (skeleton doc under -J, one stderr line in text, exit 0).

Issue #331, opened 2026-07-03
server: normal-path joins park while a completed peer holds its sockets (GT closes at TEST_END)


therealevanhenry:
From #329 r4 (NF-2, probed live during review): a peer that completes the protocol CORRECTLY — through IperfDone(16) — but then holds its sockets open parks the server's normal-path stream-task joins for the whole hold (probed: wedged at 12 s; GT exits at ~10 s on its own clock — its TEST_END arm already closed every stream socket, iperf_server_api.c:272-276).

The #325 error/terminate paths are bounded by the abort gates now, but bounding the SUCCESS path means aborting stream tasks on a clean run — a design call about graceful-join semantics (in-flight interval flush, counter finality) that shouldn't be rushed. Options: (a) join with a bounded timeout then abort; (b) close/shutdown the data sockets at TestEnd processing like GT and keep unbounded joins; (c) accept the wedge as bounded-by-peer. (b) is the GT-faithful shape.

Repro: the wire_shape holding-peer harness with final byte 16.

therealevanhenry:
Reproduced independently during #333 r2 (pre-existing on main, not a #333 regression): full exchange → DisplayResults → IperfDone(16), peer then holds ctrl+data open — GT exits on its own clock (~9.6 s, its server timer); riperf3 `-s -1` never exits (killed at 15 s). The clean round takes the non-abort join path by omission, and the data-task join parks on the held socket. After #333 this is the LAST unbounded teardown cell in the server's message surface. Note GT's mechanism here is a TIMER, not an abort — reinforcing option (b) (close data sockets at TEST_END like GT) or a bounded join, rather than extending the abort-gate pattern to the success path.

Issue #330, opened 2026-07-03
server: -J error-sink gating + full accumulated doc for generic test errors (GT parity)


therealevanhenry:
Split from #325 r1 (F4's pre-existing remainder). Two related GT divergences on the server's generic error surface, both live-verified against GT 3.21 during the #325 review:

**1. `-J` sink gating for generic test errors.** GT routes every failed-test error through `iperf_err(test, ...)`, whose json sink puts the message in the doc and keeps stderr silent under `-J` (iperf_error.c:60-61). riperf3's serve-loop generic arm prints `riperf3: error - {e}` gated only on quiet mode, not on json — so `-J` runs get a stderr line GT never emits. The ClientTerminate and (since #325) UnknownControlMessage arms are gated correctly; the generic arm and the CLI error sink are not.

**2. Accumulated doc vs skeleton.** On a failed test under `-J`, GT still emits the FULL accumulated document — populated `end` block plus the `error - `-prefixed key (main.c:173-183 → iperf_json_finish). riperf3's CLI renders `json_report::error_document` (empty `end: {}`) for error classes that don't reach the report machinery. The #325 terminate/IEMESSAGE paths prove the full-doc machinery works; the remaining error classes should ride it too where a report exists.

Related: the mid-test message loop still tolerates known-but-unexpected states (#145 arm at the TEST_RUNNING read) that GT's single `iperf_handle_message_server` switch would IEMESSAGE — same divergence class as #325's end-loop fix, one phase earlier. Worth deciding together with (1)/(2) since the surfaces overlap.

therealevanhenry:
Addendum from #329 r2 (F4): the server's END-LOOP EOF arm is also part of this surface. GT's read path (iperf_server_api.c:249-254) raises IECTRLCLOSE — "the client has unexpectedly closed the connection" — printed once (into the -J doc via iperf_err's sink), then the round ends cleanly (rc 0). riperf3's end loop `Err(PeerDisconnected) => break` reports a clean success with no message anywhere. Same for the mid-test EOF path. Fold into the generic-surface work here.

therealevanhenry:
Scoping note (PR #333 covers part of this issue — it now says Refs, not Closes):

[...]

PR #370, opened 2026-07-07
fix: an exchange-phase Err runs the end-of-round teardown (#353)

### LINKED ISSUE #353: server: an Err from the exchange phase skips the abort/join block — detached tasks can park on a holding peer
Found by PR #352 r1 (pre-existing on base, same parking class as #331).

An `Err` propagated by `exchange_results_phase` itself (the `?` at server.rs:791 — e.g. a send failure writing the server's results) returns out of `handle_one_test` BEFORE the abort/join block. The dropped JoinHandles DETACH the tokio stream tasks rather than abort them, so a hostile peer that provokes the send failure and then holds its data sockets leaks a parked TCP receiver + its fds in a persistent server until the peer relents. (The serve loop itself keeps serving — no join parks — so this is a resource leak, not a wedge.)

Fix shape: route the exchange-phase error through the same unconditional abort+join teardown (#331's block) instead of early-returning past it — e.g. capture the result, run the teardown, then return.

therealevanhenry:
Closes #353 — an `Err` from the exchange phase itself no longer skips the end-of-round teardown (the #331/#352/#356-F7 leak family, one phase later; found by PR #352's r1, pre-existing).

## The leak

`exchange_results_phase`'s `?` returned out of `handle_one_test` past the #331 **unconditional** abort/join gate. A hostile peer that provokes the exchange's own send failure — RST the ctrl inside the ~100 ms flush-grace window, before the ExchangeResults state write lands — while HOLDING its data sockets left detached receiver tasks parked in `read()` plus their fds, accumulating per round in a persistent server. (The graceful recv-fail class — IERECVRESULTS, #336 — was never affected: it routes through the report path and the gate.)

## The fix

The exchange result is captured; on `Err` the same abort+join (streams + demux) runs before propagating. Counts are frozen at `build_result_streams`, so the teardown is safe at this point — same argument as the #331 gate it mirrors.

## Verification

- Pin red-first (persistent server — the #356-F7 vacuous-pin lesson): retry-classified across the RST-vs-write race (the #345 pattern; recv-class rounds already run the gate and retry). The send-Err class asserts the held data socket sees EOF bounded while the server keeps serving. Red on main: the held socket outlived a 4 s read. Green on the fix in one round.
- Mutation from the committed tree (dirty-target-guarded): the teardown removed from the new Err arm → the pin RED.
- Full workspace suite exit 0; clippy `-D warnings`/fmt clean; xcheck 7/7.
- CI + compat 52/52 on the final head before merge.

### DIFF
--- a/riperf3-cli/tests/wire_shape.rs
+++ b/riperf3-cli/tests/wire_shape.rs
@@ -3681,3 +3681,111 @@ fn running_idle_watchdog_negative_cells() {
          RECORDED DEVIATION — GT's block-quantized signal kills this cell): {serr2:?}"
     );
 }
+
+// ---------------------------------------------------------------------------
+// #353: an Err from the exchange phase itself (e.g. the ExchangeResults
+// state-byte write failing against an RST'd ctrl) must not skip the
+// end-of-round abort/join gate — a hostile peer provoking it while
+// HOLDING its data sockets would leak detached parked receiver tasks +
+// fds in a persistent server (the #331/#352/#356-F7 family, one phase
+// later). PERSISTENT server on purpose: one-off process exit closes the
+// sockets either way (the #356 F7 vacuous-pin lesson).
+// ---------------------------------------------------------------------------
+
+/// One #353 attempt. The ctrl-RST races the exchange's first write: RST
+/// first = the send-Err class (the `?` under test); write first = the
+/// graceful IERECVRESULTS class (which runs the gate — a retry). Returns
+/// (send_class, data_read, server_alive).
+#[cfg(unix)]
+fn drive_exchange_err_hold() -> (bool, Result<usize, std::io::ErrorKind>, bool) {
+    let listener = std::net::TcpListener::bind("127.0.0.1:0").expect("bind");
+    let port = listener.local_addr().unwrap().port();
+    drop(listener);
+    let ps = port.to_string();
+    let mut server = common::ChildGuard(
+        std::process::Command::new(env!("CARGO_BIN_EXE_riperf3"))
+            .args(["-s", "-p", &ps])
+            .stdout(std::process::Stdio::piped())
+            .stderr(std::process::Stdio::piped())
+            .spawn()
+            .expect("spawn server"),
+    );
+    let _so = riperf3_test_support::drain_reader(server.0.stdout.take().unwrap());
+    let se = riperf3_test_support::drain_reader(server.0.stderr.take().unwrap());
+    std::thread::sleep(std::time::Duration::from_millis(400));
+
+    let mock = std::thread::spawn(move || {
+        use std::io::Read;
+        let cookie = [b'x'; 37];
+        let mut ctrl = std::net::TcpStream::connect(("127.0.0.1", port)).expect("ctrl");
+        ctrl.write_all(&cookie).unwrap();
+        assert_eq!(read_exact(&mut ctrl, 1)[0], 9);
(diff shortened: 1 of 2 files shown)


therealevanhenry:
## Review round record (two independent cold reviews, both APPROVE)

**r1 — APPROVE.** Headline evidence: repeat-round fd A/B on a persistent server (24 hostile exchange-fail rounds, peer holds its data sockets) — main leaks +1 fd/round monotonically with the held peer never seeing EOF; this head stays flat at baseline with EOF in ~53 ms every round, and a clean GT round completes after. Sibling-return audit: the exchange call was the only fallible early-return between `shutdown_and_flush` and the unconditional gate (`build_result_streams`/`finish_server_output`/`build_ctx_report`/`emit_final_output` are infallible) — **but** two sites one phase earlier (`start_test`, non-EOF `await_test_end`) skip the same teardown; pre-existing, live-reproduced, filed as **#372** with a drop-guard fix shape. The exchange-send stderr prose gap (raw io vs GT IESENDMESSAGE/IESENDRESULTS) filed as **#371**. Mutations: teardown-strip and abort-neuter both red on the pin.

**r2 — APPROVE.** GT source-level fidelity: GT closes stream sockets inline at TEST_END (iperf_server_api.c:271-276) so it structurally can't leak on this class; riperf3 defers to the post-exchange gate (the recorded #331 catch-up rationale), which demands teardown on every post-TestEnd exit — what this PR closes. GT keeps serving after an exchange failure (rc −1 is not < −1, main.c:169-185); live-verified this head does the same. Mode matrix live (head vs main-parent): TCP -P2 / UDP -P2 demux / UDP recycling / -R / --bidir all flat-fd with bounded EOF and a clean round after; reverse confirmed the parked class is receivers only. GT 3.21 interop: clean → SIGKILL-near-exchange ×3 → clean, server survives and re-arms, surfaces match the recorded IECTRLCLOSE convention. Windows: green by `#[cfg(unix)]` gating of the SO_LINGER pin, not by accident. SemVer: private internals only, same error value rethrown.

**r2 findings applied post-approve @6ea5631 (test-only, r2's own prescriptions — #347 precedent, no r3):**
- Positive class discriminator: `send_class` now requires an `error - ` stderr line to have landed AND no recv-results marker (the kill races the serve-loop line; the negative-only form could false-green a teardown-less mutant via a misclassified recv-class round). Line-less rounds retry.
- Bounded-close assert accepts `Ok(0) | Err(ConnectionReset)` (the abort can drop the receiver with the mock's bytes unread → RST; GT's own close on this class RSTs too). Main's failure mode stays TimedOut/WouldBlock — discrimination preserved.
- Re-proven on 6ea5631: teardown-strip mutant → pin RED in 4.53 s; restored; 10/10 free + 6/6 `taskset -c 0,1 --test-threads 2` green; clippy -D warnings clean; workspace 844/844.

**Recorded, no code change (r2 LOW/NIT):**
- The join half and demux half of the new arm are mutation-invisible (abort alone frees the TCP fds; the demux thread exits on `done` + its 500 ms read-timeout regardless). They are deliberate structural symmetry with the tested #331 gate; the join remains the only wait for spawn_blocking runners. A discriminating pin would have to be UDP-shaped and timing-based — not worth the flake surface.
- GT emits a **populated** doc (not the #330 skeleton) on the exchange-send class — noted on #371 so its fix doesn't cement the skeleton.

Compat 52/52 ran clean on 1963230; re-running on this head before merge.

Issue #372, opened 2026-07-07
server: start_test / non-EOF await_test_end errors skip the abort/join teardown — the running-phase fd-leak siblings of #353


therealevanhenry:
Found by PR #370's r1 (pre-existing — the diff doesn't touch these lines; head and main reproduce identically). Sibling of #353, one phase earlier: the same detached-parked-task fd leak via `?` sites between stream-spawn and the unconditional abort/join gate.

#370 wrapped only the exchange call (server.rs:903). Two sites in the running phase still propagate an Err out of `handle_one_test` without running the teardown:

- `start_test` at server.rs:868 — the `send_state(TestStart)` / `send_state(TestRunning)` writes (server.rs:1581/1588) `?` after streams are spawned.
- `await_test_end` at server.rs:871 — a **non-EOF** `recv_state` error (e.g. ECONNRESET, which is not mapped to `PeerDisconnected`) hits the `Err(e) => return Err(e)` arm at server.rs:1831.

Live repro on a persistent server (head 1963230): mock peer connects a data socket, then RSTs the ctrl and holds the data socket → **+1 fd/round, +20 over 20 rounds, held peer never sees EOF, server stays alive**, both timings emitting the raw `error - Connection reset by peer (os error 104)` (that prose class is #371).

Fix shape (r1's prescription, endorsed): rather than a fourth per-phase `if let Err` wrapper (whack-a-mole — this is the 4th site in the family: #331 gate, #353 exchange, now these two), a drop-guard/RAII that runs the abort/join on ANY early return after `setup_data_streams` would close `start_test`, `await_test_end`, and every future phase at once. Note the existing `_done_guard` (server.rs:858) sets `done` but does NOT wake a TCP receiver parked in `read()`, so it doesn't cover this class.

Rider for the same PR: `abort_setup_streams` (server.rs:428) is doc-scoped "setup-phase error paths (#356 r1 F7)" but now serves end-of-round sites too — broaden the doc or extract a neutrally-named shared helper (PR #370 r1 F3 nit).

Pin shape: the #370 pin's harness (persistent server, hold-the-data-socket mock, `data_read == Ok(0)` EOF assertion) retargeted at the two windows: RST-during-TestStart-send and RST-mid-TEST_RUNNING (non-EOF class).

Issue #293, opened 2026-07-02
Unify Client::run abnormal-end semantics: return the partial Report on server-terminate/relayed-error paths


therealevanhenry:
From the 0.8.0 post-release review (v0.7.4..v0.8.0, main @9e0f77e). **Breaking** — 0.9.0.

## Problem

`Client::run`'s abnormal-end paths are asymmetric, and two of them discard data the CLI user can see:

| End condition | Return | Partial `Report`? |
|---|---|---|
| Clean run | `Ok(Report)` | full |
| Signal interrupt (TERM/INT/HUP) | `Ok(Report { error: Some(..) })` (`client.rs:502`, `:709`) | returned |
| Mid-test SERVER_TERMINATE | `Err(RiperfError::ServerTerminated)` (`client.rs:678`, `:869`) | **built, printed, discarded** |
| Relayed SERVER_ERROR | `Err(RiperfError::ServerErrorRelayed)` (`client.rs:753`, `:842`) | **built, printed, discarded** |

Two problems:

1. **Lossy**: on the `Err` paths, `print_results` builds the full partial report and writes it to stdout — then `run` returns an error carrying only a message. A CLI user gets the partial data; a library caller cannot access it at all.
2. **Inconsistent**: "the test ended abnormally with partial data" is signaled two different ways depending on *which* abnormal end occurred. `Ok`-with-embedded-`error` additionally means a caller must inspect `report.error` to know whether an `Ok` run was actually clean.

## Options (all breaking)

- Attach the partial `Report` to the error variants (`ServerTerminated(Box<Report>)`, ...) — smallest surface change, keeps `Result` semantics.
- A typed outcome: `run() -> Result<RunOutcome>` with `RunOutcome { report: Report, termination: Termination }` where `Termination` is `Completed | Interrupted | ServerTerminated | ServerError(..)` — makes the interrupt path's current `Ok`-with-`error` explicit too.
- `Ok(Report)` on every path that produced a report, `Err` only when no report exists (connect/param-exchange failures), with a typed status on the report.

Whichever is chosen must preserve the CLI's exit-code faithfulness (exit 0 on TERM/INT/HUP like iperf3, non-zero on server-terminate/error) — that mapping moves into the CLI's match on the new shape.

## Notes

- Breaking: `RiperfError` variant shape and/or the `run` contract → 0.9.0 per the 0.x semver convention.
- `Server::run_once` should get the same treatment for symmetry (its client-terminated path).
[...]

PR #373, opened 2026-07-07
fix: the client aborts stream tasks before the clean-path join (#354)

### LINKED ISSUE #354: client: joins park on a socket-holding server — GT's client exits on its own clock (sibling of #331)
Found by PR #352 r2, live-proven on base AND head (client untouched by #352).

Client::run joins bare — done.store(true) + 100 ms sleep + `let _ = s.task.await` with NO abort (client.rs:486-490). A reverse client's receivers park in read().await; `done` cannot wake a parked read. Probe: mock server completes the full protocol (through reading IperfDone) then HOLDS its sockets — riperf3 client exit is peer-bound (12 s hold → 12.002 s, 16 s → 16.006 s). GT 3.21's client exits at 10.01 s on its own clock in the same scenario (net.c:75 nread_read_timeout=10 bounds the receiver select; GT also pthread_cancels + joins receiver threads post-loop, iperf_client_api.c:817-841 — the same abort-before-join shape #352 gave the server).

Fix shape: the #331 pattern on the client — unconditional abort of stream tasks at the client's join site (counts frozen by then per the same order-of-phases argument), + a holding-mock bounded-exit pin. Consider whether GT's 10 s read bound should also exist on riperf3's client data reads (net.c:75-76 class, the #339 nread_exact sibling).

therealevanhenry:
Closes #354.

## The defect

`Client::run`'s clean-path cleanup joined bare — `done.store(true)` + 100 ms grace + `let _ = s.task.await` with no abort. `done` cannot wake a stream task parked in a socket wait, so against a peer that goes quiet mid-window and then HOLDS its sockets after the round, the client's exit was **peer-bound** (r1's live A/B on the stripped fix: 12 s hold → 13.2 s exit, tracking the hold). GT's client exits on its own clock: it cancels its stream threads before joining (iperf_client_api.c:817-841).

## The fix

The #352 server-gate shape, mirrored: unconditionally `abort()` every stream task right before the join. Safety is the same order-of-phases argument:

- Senders stopped at the **data-phase** end (`done` + 100 ms teardown grace at the #159 site) — long before this join; counters are settled. (r1 mutation (c) — abort hoisted above the cleanup-site done/sleep — stays green across the CLI suite: nothing reads a counter after this site.)
- The final report was captured at DisplayResults and the results exchange is complete — the teardown can't touch any figure.
- UDP receivers are `spawn_blocking` (abort is a no-op there) and exit via `done` + their 500 ms read-timeout poll — r1 live-verified 0.595 s exits with the abort stripped, so the abort is not load-bearing for UDP, matching the #352 rationale.
- No client-side demux exists (#80's demux is server-only).

r1's live mode matrix on the head (exit measured after the mock read IperfDone): -R P1 0.170 s, -R -P4 0.283 s, --bidir 0.239 s, fwd -P2 0.299 s, UDP -R 0.552 s, parked-sender fwd 0.261 s — every cell own-clock; GT itself takes 10.1 s on the same holding mock (its bounded sync-close, see the residual below).

## The pins (reworked after r1 F1)

r1 proved the first-draft pin never parked the receiver: its mock wrote until TestEnd, so every `read` completed and the receiver landed in the **#23 `RECEIVER_DRAIN_TIMEOUT` (10 s)** — the red was `10 s > 8 s bound`, coupled to an unrelated constant (abort-stripped + bound raised to 14 s → that pin PASSED at 11.07 s). The reworked mock goes **silent/deaf at 0.3 s — before `done` lands** — so the stream task genuinely parks:

- `reverse_client_exits_bounded_while_server_holds` — receiver parked in `read().await`.
- `forward_client_exits_bounded_while_server_stops_reading` (r1 F4, the #352-r2-F2 analog) — sender parked in `write().await` on full buffers.

Both: full protocol through reading IperfDone(16), then a 25 s hold; the bounded `run_client` (8 s) is the assert, plus exit 0 + clean `iperf Done.`.

- **Red** (abort stripped from the committed tree, anchor-uniqueness-verified): both pins 8.01 s timeouts.
- **Hold-coupling discriminator** (the r1 F1 concern, re-proven on the rework): mutant vs a 12 s-hold mock exits at **13.203 s** (tracks the hold — the park, not the 10 s drain cap); restored head exits **1.305 s** on the same mock.
- **Green** on the fix; stability 5/5 free + 6/6 `taskset -c 0,1 --test-threads 2`. Cross-platform (no libc/SO_LINGER — plain socket hold), runs ungated on all native legs.
- r1 mutation (b) — join stripped, abort kept — is mutation-invisible (the #370 M3 precedent: defensive symmetry; the join is the only wait for UDP `spawn_blocking` runners, and a discriminating pin would need thread-liveness observation, flake-prone). Accepted, recorded here.

## Scope decisions + records from the r1 round

- **GT's 10 s data-read bound is NOT taken here.** The abort-at-join gives a tighter own-clock exit for this class; mid-test the client's duration timer bounds the run. The remaining unbounded client waits are the **ctrl reads** (wedged exchange → parks forever where GT self-bounds at 10 s) — now properly recorded as **#374** (r1 F2 caught that the prior "#330 residual" attribution was unbacked: #330 never carried it).
- **#375 filed** (r1 F3): abnormal-path early returns still drop JoinHandles — the client sibling of #372, same RAII prescription; coordinate with #293's arm rework.
- **Join-site residual comment added** (r1 F5): GT FINs its data sockets at DISPLAY_RESULTS before IPERF_DONE and sync-closes ctrl bounded; riperf3's FINs land ~100 ms later at the abort, ctrl closes abruptly — holding-peer-visible only.
- r1 F6 (the cleanup-site `done.store` + 100 ms is redundant — the data-phase end already did both): pre-existing, noted, untouched (100 ms of exit latency, not correctness).

## Gate

- fmt / clippy `-D warnings` / workspace suite: clean, **846 passed, 0 failed** (845 + the forward pin).
- Red/green/mutation as above; CI green on the r1 head, re-running on this one.

### DIFF
--- a/riperf3-cli/tests/wire_shape.rs
+++ b/riperf3-cli/tests/wire_shape.rs
@@ -3802,65 +3802,94 @@ fn exchange_phase_err_still_tears_down_streams() {
 
 // ---------------------------------------------------------------------------
 // #354: the clean-path join must not be peer-bound — GT's client cancels
-// its stream threads before joining (iperf_client_api.c:817-841), so it
-// exits on its own clock against a server that HOLDS its sockets after
-// the round. A bare join parks reverse receivers in read() (`done` cannot
-// wake a parked read) until the peer lets go.
+// its stream threads before joining (iperf_client_api.c:817-841). A bare
+// join parks on a stream task stuck in a socket wait (`done` cannot wake
+// a parked read or write) until the peer lets go.
 // ---------------------------------------------------------------------------
 
-/// #354: reverse client vs a socket-holding server — the mock speaks the
-/// full protocol through reading IperfDone, then HOLDS ctrl + data open.
-/// The bounded `run_client` wait IS the pin: a bare join is peer-bound
-/// (~25 s here); the abort-before-join exits on the client's own clock.
+/// #354 mock: full protocol, but the data phase goes SILENT/deaf after
+/// ~0.3 s — BEFORE the client's `done` lands — so the stream task
+/// genuinely PARKS: reverse = the writer stops writing (a receiver parked
+/// in `read().await`); forward = the drain stops reading (a sender parked
+/// in `write().await` on full buffers). Then it reads IperfDone(16) and
+/// HOLDS every socket for 25 s. The silence timing is load-bearing (r1
+/// F1): a writer that runs to TestEnd hands the receiver to the #23
+/// 10 s drain cap instead, and the pin's red couples to that unrelated
+/// constant rather than to the park.
+fn holding_mock_round(listener: std::net::TcpListener, reverse: bool) {
+    let (mut ctrl, _) = listener.accept().expect("ctrl accept");
+    read_exact(&mut ctrl, 37); // cookie
+    ctrl.write_all(&[9u8]).unwrap(); // ParamExchange
+    read_json_blob(&mut ctrl); // params
+    ctrl.write_all(&[10u8]).unwrap(); // CreateStreams
+    let (mut data, _) = listener.accept().expect("data accept");
+    read_exact(&mut data, 37); // data-stream cookie
+    ctrl.write_all(&[1u8]).unwrap(); // TestStart
+    ctrl.write_all(&[2u8]).unwrap(); // TestRunning
+                                     // ~0.3 s of live traffic, then silence. The thread holds only a
+                                     // dup'd fd — its exit does NOT close the socket while `data` lives
+                                     // on below (the hold must be real).
+    let mut data_io = data.try_clone().expect("clone data");
+    let io = std::thread::spawn(move || {
+        let mut chunk = vec![0u8; 65536];
+        let t0 = Instant::now();
+        while t0.elapsed() < Duration::from_millis(300) {
+            if reverse {
+                if data_io.write_all(&chunk).is_err() {
+                    break;
+                }
+                std::thread::sleep(Duration::from_millis(5));
+            } else if data_io.read(&mut chunk).map(|n| n == 0).unwrap_or(true) {
+                break;
+            }
+        }
+    });
+    assert_eq!(read_exact(&mut ctrl, 1)[0], 4, "TestEnd");
+    ctrl.write_all(&[13u8]).unwrap(); // ExchangeResults
+    read_json_blob(&mut ctrl); // the client's results
+    write_json_blob(&mut ctrl, MOCK_RESULTS);
+    ctrl.write_all(&[14u8]).unwrap(); // DisplayResults
+    assert_eq!(read_exact(&mut ctrl, 1)[0], 16, "IperfDone");
+    io.join().expect("data io");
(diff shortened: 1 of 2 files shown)


therealevanhenry:
## r2 round record (independent full-PR review at 9c2e6b2)

**Verdict: REQUEST-CHANGES, one MINOR finding — applied as the final commit.** Everything else verified clean:

- **Timing robustness**: silence(+0.3 s) and `done`(~t=1) share the handshake epoch (0.7 s margin); under inversion the head STAYS green (the abort kills the #23 drain too) — degradation is mutant-red attribution only, never a CI false-red. Stability 10/10 free, 6/6 two-core, 6/6 single-core, 6/6 single-core + 4 busy-loop hogs. Forward park: host max in-flight ≈36 MB vs GB/s loopback — fills in ms; vacuous pass impossible against a deaf finite-buffer peer.
- **Mock fidelity**: byte-walk matches GT's server sequence; cross-validated by running the real GT 3.21 client against a byte-identical port of the mock — clean `iperf Done.`, 11.4 s bounded exit (GT's own ctrl sync-close drain, matching the F5 residual).
- **Output identity under the abort**: 4-cell live A/B (head vs abort-stripped vs GT server): `--get-server-output -R`, `-J -P2`, `-u --bidir`, `-k 2000` — text masked-identical, JSON field-path sets identical, interval counts equal. Count-safety independently re-derived (final_report is plain owned data at DisplayResults).
- **GT anchors** all exact (cancel-before-join :817-841; data close at DISPLAY_RESULTS :563-566 before IPERF_DONE :573; ctrl sync-close :580 / net.c:877-886). The F5 residual comment's ~100 ms-later-FINs claim verified on head.
- **Records #374/#375** accurate incl. the #330-attribution correction and every cited line number.
- **Mutations**: full-strip → both pins red, hold-coupled (13.22 s tracking a 12 s hold, not the ~11.1 s drain-cap shape); order-swap (join before abort) → both red — the ORDER is pinned; senders-only/receivers-only aborts → each pin independently discriminates its direction.
- **Gate**: fmt/clippy clean, 846/846, CI 17/17 on 9c2e6b2, SemVer green (lib observes nothing but faster exits).

**r2 F1 (MINOR, fixed in the head commit)**: a `take(1)` partial teardown (abort stream 0 only) was invisible to both P1 pins — a future early-break/off-by-one in the teardown loop would regress only multi-stream runs. Fix = r2's prescription, the clean variant: `holding_mock_round` now accepts N data connections and the reverse pin drives `-P 2` (no listener-backlog reliance — both conns are accepted and go silent symmetrically; the peer blob keeps one row, which the client's merge tolerates and the pin's asserts never read). Red-proven post-commit: take(1) mutant → reverse -P2 pin red 8.01 s; full-strip → both pins red 8.02 s; restored green; 5/5 free + 6/6 two-core; workspace 846/846.

Issue #338, opened 2026-07-06
server: hang on valid-but-incomplete params + control-EOF during stream setup


therealevanhenry:
Discovered while probing #330 (2026-07-06). **Adversarial / robustness — triage for 0.9.0 vs follow-up.**

## Symptom

A one-off server (`riperf3 -s -1`) HANGS indefinitely when a client sends a *valid-JSON-but-incomplete* params blob and then closes the control socket without creating data streams. GT (iperf 3.21) proceeds past this and produces a normal doc + exit 0.

## Repro (mock client)

1. connect, send the 37-byte cookie
2. read PARAM_EXCHANGE (byte 9)
3. send a 4-byte BE length + `{"tcp":true}` (valid JSON, missing `parallel`/`len`/etc.)
4. close the control socket

- **GT**: reaches DISPLAY, emits a full `-J` doc / banner, exit 0 in ~1s.
- **riperf3**: parked >10s (no bound observed), server process never exits.

Contrast: a *malformed* params blob (bad JSON) does NOT hang — riperf3 errors out immediately (that path is #330's error-sink concern). Only the valid-but-incomplete + ctrl-EOF case hangs.

## Suspected cause

The server likely accepts the defaulted params, advances to CREATE_STREAMS, and blocks in the data-stream `accept()` without watching the control socket for EOF — so a peer that closed the control connection is never noticed. Same *class* as #331 (parking on a socket that will never make progress), one phase earlier (stream setup vs post-test join).

## Notes

- Reachable only via a hand-rolled/malformed client; a real iperf3 client never does this. So it's a robustness/DoS-adjacent gap, not a normal-operation bug.
- A bounded wait on the data-stream setup (or a select over ctrl-EOF during setup, like GT) would close it.

therealevanhenry:
Review-round addendum (from the #339 r3 probe matrix): the wedge has a second variant this issue's body doesn't yet document.

- **ctrl-EOF variant** (the original report): peer sends valid-but-incomplete params, closes the control connection, no data streams → riperf3 parks in the `setup_data_streams` accept; GT notices the EOF via its event-loop `select()` and exits in ~0.04s with the IECTRLCLOSE surface (text: `iperf3: the client has unexpectedly closed the connection`; `-J`: populated-start doc + `intervals:[]` + bare `end:{}` + error key — post-params fields like `accepted_connection`/`cookie` present, NOT the minimal skeleton).
[...]

Issue #281, opened 2026-06-24
Client pre-TestStart interrupt -J dump emits late fields where iperf3 omits them


therealevanhenry:
## Summary

The client interrupt-dump path hardcodes `reached_test_start: true` when building the `-J` report, so a `SIGINT`/`SIGTERM` that arrives **before TestStart** (e.g. while blocked on the PARAM_EXCHANGE control read, before any data stream exists) emits a document with `start.test_start` + `start.sock_bufsize`/`sndbuf_actual`/`rcvbuf_actual` and a populated `end`, whereas iperf3's `iperf_got_sigend` skips `iperf_on_test_start()` when TEST_START wasn't reached and would emit the bare-`end: {}` shape (the same shape #261 just made faithful for the upfront-refusal path).

## Context

Surfaced by the #261 cold review. The refusal path (#261) is now correctly gated on `reached_test_start`, but the interrupt arm in the client dispatch loop (`riperf3/src/client.rs`, the `wait_interrupt` select arm) still passes a constant `true`. The arm already computes `dump_secs = if test_start_millis > 0 { … } else { 0.0 }`, so the correct signal (`test_start_millis > 0`) is already in scope.

## Scope / why not #261

Different path (interrupt vs server refusal) and narrow timing window (interrupt exactly between connect and TestStart). There is an existing test (`pre_data_interrupt_dump_reports_a_zero_window`) asserting the current pre-data-interrupt shape, so the fix needs (a) GT verification that `iperf_got_sigend` really omits `test_start`/`end` pre-TestStart, and (b) realignment of that test. Kept out of #261 to hold its scope to the refusal doc.

## Fix

- Set `reached_test_start: test_start_millis > 0` in the client interrupt arm.
- Verify against GT (iperf 3.21) that a pre-TestStart SIGINT dump omits `test_start`/buffer fields and emits `end: {}`.
- Update `pre_data_interrupt_dump_reports_a_zero_window` to the GT-faithful shape.

Faithfulness, `-J` only; no wire change.

therealevanhenry:
GT verification done (iperf 3.21 @ d39cf41, live capture): client held in the post-param-exchange / pre-TestStart window by a stalling server, then SIGINT. Exit 0, one `-J` doc:

```json
{
  "start": {
    "connected": [],
    "version": "iperf 3.21",
    "system_info": "<scrubbed>",
    "timestamp": { "time": "...", "timesecs": 1783008224, "timemillisecs": 1783008224079 },
[...]

Issue #261, opened 2026-06-23
Client -J error/refusal JSON document diverges from GT (epoch-0 timestamp, extra keys, populated end block)


therealevanhenry:
> **GT** = ground truth: the pinned reference **iperf 3.21**, built from source at commit `d39cf41526626b4e5a130f115d931cd6cbdffc19`; source-line citations refer to that tree.

From the 0.7.4 cold-review ledger (unfiled finding; flagged in #249 r1 and r2, pre-existing on the #224 path).

## Problem

On the client error / refusal path, riperf3's `-J` (JSON) document diverges from GT's error-document skeleton:

- **Epoch-zero timestamp:** riperf3 emits a `timestamp` of epoch 0 instead of GT's shape.
- **Extra keys:** riperf3 includes `sock_bufsize` / `test_start` keys that GT does not emit on the error document.
- **Populated `end` block:** riperf3 emits a populated `end` object where GT emits `{}` (empty) for an error/refusal.

## Fix direction

Shape the error/refusal `-J` document to GT's skeleton: empty `end` block, no spurious `sock_bufsize`/`test_start`, and a correct (or omitted) timestamp. Audit GT's `iperf_json_finish` / error path for the exact emitted key set.

Compat / faithfulness.

Issue #271, opened 2026-06-23
GT old-peer all-omitted posture is unmirrored (#24 class)


therealevanhenry:
> **GT** = ground truth: the pinned reference **iperf 3.21**, built from source at commit `d39cf41526626b4e5a130f115d931cd6cbdffc19`; source-line citations (e.g. `iperf_api.c:2945-2949`) refer to that tree.

From the 0.7.4 cold-review ledger (referenced as tracked-separately in #255 / #256 comments; #24 class).

## Problem

GT applies an **"all-omitted" posture for old peers** (`iperf_api.c:2945-2949`) — when the peer is an older iperf3 that does not exchange certain counts, GT treats the affected figures as all-omitted. riperf3 does not mirror this old-peer handling, so a run against an old peer can diverge from how GT would render the same exchange.

## Fix direction

Mirror GT's old-peer all-omitted posture for the affected aggregates (interacts with the #31 / #214 omit-netting scope and the #256 datagram-counter work). Determine the peer-version gate GT uses and reproduce it.

Compat / faithfulness.

therealevanhenry:
Scope decision (Evan): GT's old-peer substitution renders self-inconsistent figures — real loss swallowed on no-omit runs, the `-1` sentinel leaking `+1` into the stream count while the sum skips it (the same document disagrees with itself), and a `lost_percent` denominator collapsing to 0 beside nonzero `lost_packets`. We are NOT mirroring those; riperf3 nets absent `omitted_*` with the best local estimate (own omitted sent/received count per stream direction; error totals never netted) — exact without `-O`, consistent across every surface with it. The upstream bug is filed as esnet/iperf#2055 and the deviation record lives on `protocol::resolve_peer_omitted`. What riperf3 keeps from GT: the Option-ized absent-vs-zero decode and the both-or-neither IERECVRESULTS guard.

PR #376, opened 2026-07-08
fix: the server -J start block emits the bufsize trio at GT's positions 4-6 (#355)

### LINKED ISSUE #355: server -J start block: sock_bufsize/sndbuf_actual/rcvbuf_actual serialize late; GT's server emits them before timestamp
Found while building #338's setup-phase doc. GT's SERVER-role `start` key order (normal run AND the #338 CREATE_STREAMS doc, live 3.21):

`connected, version, system_info, sock_bufsize, sndbuf_actual, rcvbuf_actual, timestamp, accepted_connection, cookie, tcp_mss_default, target_bitrate, fq_rate[, test_start]`

riperf3's server emits the three bufsize keys AFTER fq_rate (the shared Start serializer's Started-stage tail, json_report.rs:308-318). GT's server adds them to json_start at iperf_tcp_listen time (iperf_tcp.c:373-377) — insertion order puts them at positions 4-6 on the server role. JSON-semantically identical; byte-order divergence only (the #300 raw-key-order class).

Fix shape: the Start serializer already knows the role (accepted_connection vs connecting_to) — emit the bufsize trio early on the server role, keep the client order (verify the client's GT order first). Raw key-order pin per #300's extractor precedent.

therealevanhenry:
Closes #355.

## The defect

GT's TCP server inserts the bufsize trio (`sock_bufsize`, `sndbuf_actual`, `rcvbuf_actual`) into `json_start` at listen time (iperf_tcp.c:372-375), so its SERVER-role `start` block carries them at positions 4-6, before `timestamp`. The client takes them at TestStart, after `fq_rate`. riperf3's shared Start serializer emitted the client-derived order on both roles. Byte-order divergence only (the #300 raw-key-order class) — the key/value SET is identical, so this is a JSON-semantic no-op.

Live-probed orders (3.21 loopback, raw depth-1 extraction):

- GT server: `connected, version, system_info, sock_bufsize, sndbuf_actual, rcvbuf_actual, timestamp, accepted_connection, cookie, tcp_mss_default, target_bitrate, fq_rate, test_start`
- GT client: `connected, version, system_info, timestamp, connecting_to, cookie, tcp_mss_default, target_bitrate, fq_rate, sock_bufsize, sndbuf_actual, rcvbuf_actual, test_start` — **already byte-identical to riperf3's client**, so the client is untouched.

## The fix

The hand-written Start serializer (the #281 stage-gated impl) now emits the trio right after `system_info` on the server role (`accepted_connection.is_some()`), under the same `stage == Started` gate (#261 semantics unchanged — the trio still never appears in pre-TestStart docs; the #338/#356 setup-phase docs have their own builder that already carried GT's server order and explicitly deferred this shared-serializer gap to #355). The late slot skips the trio on the server role only.

## Verification

- **Pin**: `server_role_start_key_order_is_gt_server_order` (raw-string key extractor, the #300/#281-r1-F2 precedent) — **red** on the shared order (trio after `fq_rate`), green on the fix. The pre-fix red is the mutation proof: the diff IS the reorder, so reverting it is exactly the proven-red state. The existing client-order pin (`manual_serializer_key_order_is_pinned`) passes untouched — the client wire shape is frozen.
- **Live loopback** post-fix: riperf3 server and client `start` key orders both byte-order-match GT 3.21 (extractor compare, both roles).
- **Gate**: fmt / clippy `-D warnings` clean; workspace **847 passed / 0 failed**. One unrelated flake in the first full run (`exchange_eof_prints_warning_line_and_summary_in_text`, mock UnexpectedEof under parallel suite load — the #349-F4 port-race/load family; 5/5 green isolated, full wire_shape binary green on rerun, untouched surface): disclosed, no change.

**Round 2 (commit 2, r1 F1):** the gate above is TCP-only — GT's UDP server keeps the trio in the late slot (iperf_udp_buffercheck at CREATE_STREAMS, iperf_udp.c:440-452); `Start` carries an explicit `tcp` field from the single build site. UDP-server pin red-first; #377/#378 filed for the pre-existing rate-set-dup and --get-server-output-embed divergences. Full round records in the PR comments. r2: APPROVE (24-cell live GT matrix, 4 mutations red, 848/848).

### DIFF
--- a/riperf3/src/json_report.rs
+++ b/riperf3/src/json_report.rs
@@ -447,6 +447,23 @@ impl Serialize for Start {
         m.serialize_entry("connected", &self.connected)?;
         m.serialize_entry("version", &self.version)?;
         m.serialize_entry("system_info", &self.system_info)?;
+        // #355: the SERVER inserts the bufsize trio at listen time
+        // (iperf_tcp.c:373-377) — positions 4-6, BEFORE timestamp; the
+        // client takes them at TestStart, after fq_rate (both live-probed
+        // 3.21). Same #261 stage gate either way; byte order is the only
+        // role difference.
+        let server_role = self.accepted_connection.is_some();
+        if server_role && self.stage == StartStage::Started {
+            if let Some(v) = &self.sock_bufsize {
+                m.serialize_entry("sock_bufsize", v)?;
+            }
+            if let Some(v) = &self.sndbuf_actual {
+                m.serialize_entry("sndbuf_actual", v)?;
+            }
+            if let Some(v) = &self.rcvbuf_actual {
+                m.serialize_entry("rcvbuf_actual", v)?;
+            }
+        }
         if self.stage != StartStage::Connecting {
             // GT stage 1 — stamped at on_connect (end of PARAM_EXCHANGE).
             m.serialize_entry("timestamp", &self.timestamp)?;
@@ -468,14 +485,18 @@ impl Serialize for Start {
         }
         if self.stage == StartStage::Started {
             // GT stage 2 — the #261 late fields, stamped at TestStart.
-            if let Some(v) = &self.sock_bufsize {
-                m.serialize_entry("sock_bufsize", v)?;
-            }
-            if let Some(v) = &self.sndbuf_actual {
-                m.serialize_entry("sndbuf_actual", v)?;
-            }
-            if let Some(v) = &self.rcvbuf_actual {
-                m.serialize_entry("rcvbuf_actual", v)?;
+            // The bufsize trio already went out at positions 4-6 on the
+            // server role (#355).
+            if !server_role {
+                if let Some(v) = &self.sock_bufsize {
+                    m.serialize_entry("sock_bufsize", v)?;
+                }
+                if let Some(v) = &self.sndbuf_actual {
+                    m.serialize_entry("sndbuf_actual", v)?;
+                }
+                if let Some(v) = &self.rcvbuf_actual {
+                    m.serialize_entry("rcvbuf_actual", v)?;
+                }
             }
             if let Some(v) = &self.test_start {
                 m.serialize_entry("test_start", v)?;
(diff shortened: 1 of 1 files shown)


therealevanhenry:
## r1 round record

**r1: REQUEST-CHANGES — F1 fixed @a76c547; F2/F3 filed; N1/N2 applied.**

- **F1 (MAJOR, fixed):** the role gate was protocol-blind — only GT's TCP listener inserts the trio early (iperf_tcp.c:372-375); the UDP server's trio lands at CREATE_STREAMS via `iperf_udp_buffercheck` (iperf_udp.c:440-452, after on_connect), i.e. the late/client slot — so the first commit regressed the previously-matching UDP server cell (live-proven, incl. --json-stream and the GT-client cross pairing). Fix per r1's prescription: `Start` carries an explicit `pub(crate) tcp` (set from `ReportInput.protocol` at the single build site — the `stage` precedent, no `tcp_mss_default.is_some()` inference); early emit = `tcp && accepted_connection && Started`; late-slot skip under the same predicate. New pin `udp_server_role_start_key_order_keeps_the_late_trio`, red-first on the blind gate (that red is also the mutation proof for dropping the `tcp` conjunct). Live loopback post-fix: UDP server matches GT modulo the recorded #377 dup; TCP server, TCP client re-verified matching.
- **F2 (pre-existing) → #377 filed:** GT's server emits `target_bitrate` twice on rate-set runs (iperf_api.c:2662 + :1029), shifting the trio to 5-7 on rate-set TCP; riperf3 emits once. The #355 pins deliberately bake in the zero-rate shape; #377 owns the rate-set delta (likely record-deviation + upstream per the #271 ethos).
- **F3 (pre-existing) → #378 filed:** the --get-server-output embed round-trips through `serde_json::Value`, alphabetizing the entire attached doc — the #355 fix is invisible on that path until #378 lands.
- **N1:** the three verbatim raw-key extractor closures collapsed into one `raw_keys` test helper (now 4 call sites).
- **N2:** GT cites corrected to iperf_tcp.c:372-375 everywhere.

r1's mutation matrix on the first commit (role-gate revert → server pin red; both-roles-early → client pin red; late-slot-skip dropped → server pin red on the DOUBLED key — the raw extractor catches duplicates) all stands on this head; the added `tcp`-conjunct mutant is the UDP pin's proven red.

Gate on a76c547: fmt/clippy clean, workspace **848/848** (the UDP pin is +1), live UDP+TCP GT compare clean.

PR #379, opened 2026-07-08
fix: every post-setup exit runs the stream teardown, both roles (#372, #375)

### LINKED ISSUE #372: server: start_test / non-EOF await_test_end errors skip the abort/join teardown — the running-phase fd-leak siblings of #353

### LINKED ISSUE #375: client: abnormal-path early returns drop stream JoinHandles — detached parked tasks leak for library consumers (client sibling of #372)
Found by PR #373's r1 (F3). The client sibling of #372: every early return from `Client::run` drops the `ctx.streams` JoinHandles without abort/join — the tasks detach, and a task parked in a socket wait (`read().await`/`write().await` against a holding peer) leaks with its fd until the peer releases. `_done_guard` (client.rs:386 @2310eec) sets `done` but cannot wake a parked read — the same caveat #372 records for the server's guard.

Sites (head 2310eec): the returns at client.rs:410, 412, 415, 439, 448, 455, 465 plus the `?` propagations at :421/:426/:435/:437/:443. The clean path is fixed by #373 (abort-before-join); the abnormal paths bypass that site entirely.

Impact: library consumers (`run()` in-process) — a CLI process exits and the OS reaps. Same blast radius argument as #372.

Fix shape: #372's prescription verbatim — a drop-guard/RAII owning the abort/join teardown so ANY exit path after stream spawn runs it, rather than per-arm wrappers (whack-a-mole; this family is now 5 sites across both roles). Note #293 (0.9.0 milestone) reworks the abnormal-end RETURN SHAPES (partial Report) but not teardown — if #293 restructures these arms, land the guard there or immediately after; coordinate the two rather than editing the same arms twice.

therealevanhenry:
Closes #372. Closes #375.

## The defect (one family, both roles)

The teardown-skip fd leak, sites 4 and 5 (#331 gate → #353 exchange → these): a `?` between stream-spawn and the unconditional abort/join gate detaches tasks parked in `read()`/`write().await` — a parked await cannot see `done`, so each leaks with its fd until the peer lets go.

- **Server (#372):** `start_test`'s TestStart/TestRunning sends and `await_test_end`'s non-EOF recv arm (e.g. ECONNRESET, not mapped to the EOF class) propagated out of `handle_one_test` past the gate. Live repro on the issue: +1 fd/round on a persistent server, held peer never sees EOF.
- **Client (#375):** every abnormal-path early return from `run()` dropped the stream JoinHandles — only the clean path had the #354 abort-before-join. Blast radius = library consumers (`run()` in-process); a CLI process exit closes the fds and masked the class.

## The fix

No 4th/5th per-arm wrapper. Each role's fallible region now runs inside an inner async block whose outcome is checked only **after** the one unconditional teardown, so every `?` and early return falls through the gate (full JOIN semantics — the async-RAII alternative can't await joins in Drop):

- **Server:** the whole TestStart→final-output pipeline is the block; the gate follows; then `outcome?` and the existing post-gate arms. The #353 arm's duplicated teardown folds back to a plain `?`, and `abort_setup_streams` reverts to setup-phase-only call sites (resolves the #370 r1 F3 doc rider — its doc is accurate again). The gate now sets `done` itself: on the new Err paths `shutdown_and_flush` never ran, and a `spawn_blocking` UDP runner ignores `abort()` and exits via `done` + its 500 ms read-timeout poll — joining it with `done` unset would hang. Idempotent on the clean path.
- **Client:** the dispatch loop is the block (`Ok(Some(report))` = early-exit rounds that already rendered — the interrupt dumps; `Ok(None)` = the clean IperfDone break). The teardown is the #354 clean-path sequence, now unconditional; the 100 ms catch-up grace is skipped when no streams were spawned, so pre-stream error paths (connect, ParamExchange) add no latency. Error paths with live streams gain ~100 ms + the join before the error surfaces — GT-like (its cleanup paths run `iperf_client_end`, which cancels and joins the stream threads).

Diff note: the block re-indents both regions; review with whitespace-ignoring diff to see the real changes.

## Verification

**Pins, red-first on the unguarded build** (each red = the held-socket read timing out, i.e. the leak itself; the park is proven by silence/hold before the kill — the #354 r1 F1 hold-coupling lesson):

- `running_phase_err_at_test_start_still_tears_down_streams` / `..._mid_running_...` (wire_shape): persistent server, -P 2 holding mock, SO_LINGER(0) ctrl RST in each window; setup completion proven by TestStart(1) before the RST (a setup-window landing runs its own teardown and would false-green); positive stderr classification per the #370 r2 lesson (clean-EOF and exchange classes run the gate and retry).
- `client_error_return_tears_down_stream_tasks` (new lib test `riperf3/tests/teardown.rs`): **in-process** reverse -P 2 client on an explicit runtime that stays alive through the asserts (dropping it would reap the leaked tasks and mask the class); mock goes silent, FINs the ctrl mid-running, then reads the held data sockets.

**Mutation table** (harness refuses dirty targets, anchor-uniqueness checked, exact restore verified): M1 server gate-bypass (`outcome?` hoisted before the gate = main-equivalent) → both #372 pins red; M1b same mutant → the folded #353 exchange pin still reds; M2 server abort/join `take(1)` → red via the second held socket; M3 client gate-bypass → red; M4 client `take(1)` → red; M5 client abort-stripped (bare join) → red via the 8 s bound (the join parks peer-bound).

Not pin-covered (disclosed): dropping the gate's new `done.store` only bites the UDP `spawn_blocking` join-hang class — a TCP pin can't see it and a UDP window pin would need the full UDP mock stack; the store is comment-guarded at the site with the rationale.

**Gate:** fmt / clippy `-D warnings` clean; workspace **851 passed / 0 failed** (848 + the 3 pins).

## Coordination

#293 (0.9.0 milestone) reworks the client's abnormal-end **return shapes** (partial Report) but not teardown; it lands after this and adapts inside the block — recorded on #375.

**Rounds:** r1 APPROVE (5 non-blocking: #380 + #381 filed; the done-gated grace, the reporter-asymmetry record, and the block re-indent applied @73d1838). r2 APPROVE (cold, no blockers; #381 extended with the client `create_streams` twin). Full round records in the PR comments. Merge-head evidence on 73d1838: CI 17/17, mutation table 6/6 red re-run post-fixes, sandbox compat matrix 52/52 PASS.

### DIFF
--- a/riperf3-cli/tests/wire_shape.rs
+++ b/riperf3-cli/tests/wire_shape.rs
@@ -3915,3 +3915,154 @@ fn forward_client_exits_bounded_while_server_stops_reading() {
         client.stdout
     );
 }
+
+// ---------------------------------------------------------------------------
+// #372: a start_test / await_test_end Err must not skip the end-of-round
+// abort/join gate — the RUNNING-phase siblings of #353 (the same
+// detached-parked-task fd leak, one and two phases earlier). PERSISTENT
+// server on purpose: one-off process exit closes the sockets either way
+// (the #356 F7 vacuous-pin lesson).
+// ---------------------------------------------------------------------------
+
+/// Params for the #372 drives: parallel 2, so a PARTIAL teardown (abort
+/// stream 0 only) still reds (the #354 r2 F1 lesson).
+#[cfg(unix)]
+const MOCK_PARAMS_P2: &str = r#"{"tcp":true,"omit":0,"time":1,"num":0,"blockcount":0,"parallel":2,"len":131072,"pacing_timer":1000,"client_version":"riperf3 0.0.0"}"#;
+
+/// One #372 attempt. The mock completes setup — proven complete by
+/// TestStart(1) arriving on ctrl, so the RST below cannot land in a
+/// setup-phase arm (those run their own teardown and would false-green) —
+/// optionally rides TEST_RUNNING for ~0.3 s of traffic, then
+/// SO_LINGER(0)-RSTs the ctrl while HOLDING both data sockets. The RST
+/// surfaces as a send-Err inside start_test or a non-EOF recv Err inside
+/// await_test_end; both `?` past the gate on an unguarded build.
+/// Classification (the #370 r2 positive-classification lesson): a valid
+/// round printed an `error - ` line — the clean-EOF class (which runs the
+/// whole normal flow, gate included) prints its line WITHOUT the prefix,
+/// and a line-less round (the kill racing stderr) just retries. The
+/// exchange class ("unable to receive results") also runs the gate (#353)
+/// and is excluded. Returns (valid_class, data_reads, server_alive).
+#[cfg(unix)]
+fn drive_running_err_hold(
+    mid_running: bool,
+) -> (bool, Vec<Result<usize, std::io::ErrorKind>>, bool) {
+    let listener = std::net::TcpListener::bind("127.0.0.1:0").expect("bind");
+    let port = listener.local_addr().unwrap().port();
+    drop(listener);
+    let ps = port.to_string();
+    let mut server = common::ChildGuard(
+        std::process::Command::new(env!("CARGO_BIN_EXE_riperf3"))
+            .args(["-s", "-p", &ps])
+            .stdout(std::process::Stdio::piped())
+            .stderr(std::process::Stdio::piped())
+            .spawn()
+            .expect("spawn server"),
+    );
+    let _so = riperf3_test_support::drain_reader(server.0.stdout.take().unwrap());
+    let se = riperf3_test_support::drain_reader(server.0.stderr.take().unwrap());
+    std::thread::sleep(std::time::Duration::from_millis(400));
+
+    let mock = std::thread::spawn(move || {
+        use std::io::Read;
+        let cookie = [b'x'; 37];
+        let mut ctrl = std::net::TcpStream::connect(("127.0.0.1", port)).expect("ctrl");
+        ctrl.write_all(&cookie).unwrap();
+        assert_eq!(read_exact(&mut ctrl, 1)[0], 9);
+        write_json_blob(&mut ctrl, MOCK_PARAMS_P2);
+        assert_eq!(read_exact(&mut ctrl, 1)[0], 10);
+        let mut datas = Vec::new();
+        for _ in 0..2 {
+            let mut data = std::net::TcpStream::connect(("127.0.0.1", port)).expect("data");
+            data.write_all(&cookie).unwrap();
+            datas.push(data);
+        }
+        assert_eq!(read_exact(&mut ctrl, 1)[0], 1, "TestStart");
+        if mid_running {
+            assert_eq!(read_exact(&mut ctrl, 1)[0], 2, "TestRunning");
+            for data in &mut datas {
+                data.write_all(&[0u8; 65536]).unwrap();
+            }
+            // Let the receivers drain the burst and PARK in read().await —
+            // the hold below must hold parked tasks, not live ones.
+            std::thread::sleep(Duration::from_millis(300));
+        }
+        let linger = libc::linger {
+            l_onoff: 1,
+            l_linger: 0,
+        };
+        let rc = unsafe {
+            use std::os::fd::AsRawFd;
+            libc::setsockopt(
+                ctrl.as_raw_fd(),
+                libc::SOL_SOCKET,
+                libc::SO_LINGER,
+                std::ptr::from_ref(&linger).cast(),
+                std::mem::size_of::<libc::linger>() as libc::socklen_t,
+            )
+        };
+        assert_eq!(rc, 0, "SO_LINGER setsockopt failed");
+        drop(ctrl);
+        let mut reads = Vec::new();
+        for data in &mut datas {
+            data.set_read_timeout(Some(std::time::Duration::from_secs(4)))
+                .expect("set_read_timeout");
+            let mut buf = [0u8; 16];
+            reads.push(data.read(&mut buf).map_err(|e| e.kind()));
+        }
+        drop(datas);
+        reads
+    });
+
+    let reads = mock.join().expect("mock");
+    let alive = server.0.try_wait().expect("try_wait").is_none();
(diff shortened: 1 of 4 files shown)


therealevanhenry:
## r1 round record (cold, worktree, @1972e4b)

**r1: APPROVE — 5 non-blocking findings; F1/F4 filed, F2/F3/F5 applied @73d1838.**

- **F1 (minor) → #380 filed:** the block pattern covers every *return* but not a *dropped* future — a library consumer cancelling `run()` via `timeout`/`select!` leaks detached parked tasks on the cancel branch (pre-existing; an abort-only Drop guard would close the TCP half).
- **F2 (minor, recorded in-code):** the server gate deliberately does not reap `ctx.interval_handle` — on Err paths the reporter self-terminates detached via `done` (bounded ~1 s tick + 2 s wait), prints nothing post-done, holds no fd; an abort could kill a text-mode emit mid-line. Comment added at the gate, including the done-store-BEFORE-abort/recycled-raw_fd rationale.
- **F3 (minor, applied):** the client gate's 100 ms grace is now owed-checked via `done.swap` — run_test's end-of-data cleanup already slept the same grace on every path through it (clean break + mid-running arms); the grace remains where the gate is the first to stop the streams (errors between CreateStreams and the data phase). The #352 RESIDUAL record updated (data FINs land at the abort, one state past GT's close point — closer to GT than the old +100 ms).
- **F4 (minor) → #381 filed:** setup-phase residual `?` sites past `abort_setup_streams` (accept/EMFILE, recv_cookie, sockopt, capture_stream_meta, dispatch_setup_ctrl_byte) — same family one phase earlier, out of this PR's post-setup scope.
- **F5 (nit, applied):** the client dispatch loop re-indented to the block depth rustfmt skipped.

r1 verification: full semantic audit of both restructures (order-of-effects vs main: post-gate arm precedence, clean-path residue order, #353 fold exactness, done idempotence); UDP spawn_blocking join-hang claim verified in stream.rs (500 ms SO_RCVTIMEO + done poll); pin classification audited against every serve-loop stderr arm (no constructible false-green class); gate 851/851 twice, zero flakes; M1/M3 mutants re-proven by hand (red via the held-socket leak itself).

therealevanhenry:
## r2 round record (cold, worktree, @73d1838)

**r2: APPROVE — no blockers; 1 tracking-issue finding, applied to #381.**

- **F1 (minor, applied):** the client twin of #381 was unfiled — `create_streams` (client.rs:1195-1340) spawns per-iteration tasks into a local vec; a later iteration's `?` drops the vec detached, and `ctx.streams` is only assigned on Ok, so the new gate reaps nothing on that path. #381 retitled to cover both roles + scope comment added (no code change in this PR; pre-existing, setup-window family).

r2 verification (independent, cold): whitespace-normalized + comment-stripped token diff vs main — the code delta is exactly the claimed restructure, nothing smuggled under the re-indent; behavior-drift audit clean (the #353 fold is a strict subset of the gate; `outcome?` precedence matches main; client clean-path residue order preserved; the malicious-server DisplayResults-without-data corner keeps main's timing via the swap); the `done.swap` grace enumerated per path (run_test's own store+grace covers every path through it — no `?` precedes it; dropping main's second clean-path sleep is count- and wire-safe: FINs stay one state later than GT's DISPLAY_RESULTS close; the pre-data grace is live, not dead weight — senders are past the start barrier from on_test_start); the reporter non-reap rationale verified in reporter.rs (tick checks done before any emit; the only post-done emit needs reporter_end.finish(), never fired on Err paths; per-round Arcs prevent cross-round pollution); pin classification re-derived from the serve-loop arms (SendControlFailed is negotiate-only, so a failed start_test send stays in the generic `error - ` class; idle/watchdog classes unreachable at test timing); the newly covered rate-limit/watchdog relay writes (`send_server_error?` at server.rs:1880/:1913) confirmed inside the block; gate 851/851, no flakes; M1 + M3 mutants re-proven by hand.

Merge-head evidence (all on 73d1838): CI 17/17 green (SHA-keyed watch); mutation table 6/6 red re-run after the r1 fixes (M1 server gate-bypass, M1b folded-#353-pin cross-check, M2 server take(1), M3 client gate-bypass re-anchored to the swap, M4 client take(1), M5 client abort-strip); sandbox deploy + compat matrix **52/52 PASS**.

Issue #248, opened 2026-06-11
client SERVER_ERROR relay: missing GT's trailing perr ': ' suffix on the duration-expired line


therealevanhenry:
> **GT** = ground truth: the pinned reference **iperf 3.21**, built from source at commit `d39cf41526626b4e5a130f115d931cd6cbdffc19`, that riperf3's wire/behavior faithfulness is captured from and verified against (source-line citations like `iperf_api.c:4288` refer to that tree).

Found by PR #247's r1 cold review (od-verified byte-exact, pre-existing): on the `--server-max-duration` timer relay, GT's client prints the strerror line with a trailing perr suffix —

```
iperf3: SERVER ERROR - server test duration expired: \n
```

— the `: ` coming from iperf_error.c's errno==0 perr path (~:564-565). riperf3 prints `...expired` with no `: `. The same string is pinned by `self_terminate.rs::max_duration_timer_relays_expired`, so fixing the line means updating that pin (and `max_duration_timer_survives_rate_ticks`).

Open question to resolve when fixing: which IE codes take the perr branch (and thus get the suffix) — the bitrate-limit relay (IETOTALRATE=27) was NOT observed with a trailing `: `, so this is per-code, not blanket. Map the perr-class codes from iperf_error.c before changing any shared formatting. Error-string mapping territory, sibling of the failed-exchange stderr string noted in #246.

Issue #225, opened 2026-06-11
Client -J on peer-terminate errors prints to stderr; iperf3 keeps stderr empty (doc-only)


therealevanhenry:
From #223's round-2 review (non-blocking): when the SERVER terminates mid-test, a riperf3 `-J` client renders the doc with the error key correctly but ALSO prints `riperf3: error - the server has terminated` to stderr via the generic CLI Err path — iperf3 keeps stderr empty (the error rides only the doc). Pre-existing on main (the #218 sink dispatch routes run()-returned errors, but ServerTerminated's doc was already printed by the lib, so the CLI's doc print duplicates… actually the CLI prints the error STDERR line through the non-JSON fallback? — needs a trace; the fix is likely suppressing the generic print when the lib already rendered the error into a document). Same family as #224; post-0.7.3.

PR #382, opened 2026-07-08
fix: the client's results read gets GT's Nread bounds + IERECVRESULTS surface (#374)

### LINKED ISSUE #374: client: ctrl reads are unbounded — GT bounds every Nread at 10s/30s (client-side parity of the #339 server bounds)
Found by PR #373's r1 (F2). The client's ctrl reads are unbounded where GT bounds every `Nread` at 10 s idle / 30 s overall (net.c:75-76) and sync-closes with a bounded drain-to-EOF (net.c:877-886).

Live proof (head @2310eec): a mock server that sends ExchangeResults, reads the client's results blob, then never answers → the riperf3 client parks in `recv_results` until the peer releases (16.36 s observed on a 15 s wedge; forever if never released). GT 3.21 on the same wedge self-bounds: `warning: Failed to read JSON data size - read returned 0` at exactly the 10 s Nread bound, then sync-close ≤10 s more.

**Record correction:** PR #339's body called client read bounds "the recorded #330 residual", but #330 (closed 2026-07-06) never carried that item — its body/comments cover -J sink gating, doc bodies, and client *wording* parity only. This issue is the record now.

Fix shape: the #339 machinery (`nread_step` 10 s-idle/30 s-overall bounds, `nread_exact`, `json_read_bounded`) applied to the client's ctrl reads — `recv_state` waits, `recv_results`, and the exchange reads. Surface on expiry: GT's IERECVRESULTS class at the exchange (`unable to receive results`), state-read classes elsewhere — live-probe GT per window when implementing. Warning-line parity (`Nread_json` warnings) is the same family; the #336/#345 honest-errno conventions apply.

therealevanhenry:
Closes #374.

## The defect

The client's exchange reads were unbounded: a server that sends ExchangeResults, reads the client's blob, and then wedges parked the client until the peer released (the issue's live repro: 16+ s on a 15 s wedge, forever on a hold). GT self-bounds via its Nread 10 s idle / 30 s overall budgets (net.c:75-76) and errexits IERECVRESULTS.

## GT live probes (3.21) — they narrowed the issue's scope

- **State-byte WAITS are UNBOUNDED in GT.** Silent-after-accept, silent-after-TestEnd (no ExchangeResults state), and silent-before-DisplayResults wedges all exceeded a 45 s cap. riperf3's unbounded `recv_state` waits are already faithful and stay — the issue's "recv_state waits" item is probe-corrected (comment added on #374).
- **The exchange read IS bounded**: wedge-hold → `warning: Failed to read JSON data size - read returned 0; errno=0` at the 10 s idle bound, then `iperf3: error - unable to receive results:` (the #248 dangling errno-0 perr form), exit 1 at ~21.6 s (the extra ~10 s is GT's sync-close drain-to-EOF — the #354-recorded residual riperf3 deliberately lacks, so riperf3 exits ~10 s sooner on the same wedge).
- Mid-blob wedge → the `expected 500 bytes but received 5; errno=0` warning arm, same errexit.
- EOF at the results slot → same warning; the error tail appends a **stale errno's strerror** (`: Transport endpoint is not connected`) — the #330-recorded GT deviation, not mirrored; the errno-0 dangling form is emitted instead (the server-line precedent).
- **-J doc shape**: populated start, bare `end: {}`, error value `"unable to receive results: "` (no `error - ` prefix in-doc), warning still on stderr, exit 1. Intervals: GT shows collected-so-far, whose closing interval is still pending display (-t 1 → `[]`, -t 3 → 2 rows); riperf3's reporter flushes the closing interval before the exchange, so its count runs one ahead — the #55 flush-order drift, RECORDED here and deliberately un-pinned.

## The fix

- `recv_results_server` (bounded, warning-parity) is now `recv_results` for **both roles** — GT's `get_results` (iperf_api.c:2801) is called by both roles (:2400/:2404), so the five Nread_json warning arms and the bounds are shared. The plain client `recv_results`/`json_read` are deleted (their tests retarget `json_read_bounded`).
- The client's fail arm (`on_recv_results_failed`, the on_ctrl_lost shape): JSON sinks get the doc (bare end, dangling error value), text gets no dump (GT prints no summary on this class — probed), and the class errexits `RecvResultsFailed`.
- CLI: the class joins the #225 single-doc suppress list; the text line renders GT's perr tail (`riperf3: error - unable to receive results: `).
- **r1 F1 (blocker, commit 2):** `read_server_error_payload` — the SERVER_ERROR relay payload — was a bare unbounded `read_exact`, and unlike the results read it races NO interrupt arm, so the park was **signal-immune** (r1's live probe: SIGTERM did not free it). GT bounds these two Nreads (iperf_client_api.c:393-401), exiting a hold at ~30 s with a garbage-errno line read from the UNINITIALIZED short-read buffer. The read now runs on the house Nread budgets; the deterministic `"server error"` None-fallback is the recorded deviation (GT's uninitialized-memory line is not mirrored; its rc<0 IECTRLREAD arm folds into the same fallback — recorded in-code), and the 10 s idle bound fires first on a fully-silent hold (~10 s vs GT's ~30 s — the nread_step per-step-idle recorded-deviation class).

## Verification

**Pins, red-first on the unbounded build** (the wedge pins parked past the 25 s harness bound — the leak is the red; the -J pin rendered the CLI's skeleton `"early eof"` doc with no warning):

- `client_exchange_wedge_self_bounds_with_gt_warning` — wedge-hold: exit 1, the size-read warning, the dangling line, NO summary separator.
- `client_exchange_mid_blob_wedge_warns_partial_count` — 500-byte promise + 5 bytes + hold: the expected/received arm.
- `client_exchange_eof_takes_gt_recv_results_shape_in_json` — EOF, -J: doc error value + bare end + warning-only stderr + exit 1.
- `client_server_error_payload_wedge_self_bounds` — SERVER_ERROR byte then hold: parked past the 25 s harness bound red; self-bounds at ~10 s with the fallback line + exit 1 on the fix.

**Mutation table** (harness refuses dirty targets, anchor-uniqueness, exact restore): M1 nread_step unbounded (idle+overall arms stripped) → wedge pin red via the park; M2 size-read warning suppressed → red; M3 expected/received counts zeroed → red; M4 client doc emission stripped → -J pin red; M5 CLI suppress-list entry dropped (second doc concatenates) → red; M6 perr tail dropped → red; M7 payload-read bound reverted to the bare read_exact → red via the park (7/7 re-run on the r1-fixed head).

**Gate:** fmt / clippy `-D warnings` clean; workspace **855 passed / 0 failed** (851 + the 4 pins). One full-suite run flaked two pre-existing wire_shape tests (`cookie_failure_wire_back_is_server_error_ierecvcookie`, `clean_finish_reverse_exits_bounded_while_peer_holds` — mock EOF under parallel load, the #349-F4 family): green isolated, green ×3 in-binary, green on two full reruns; disclosed, no change.

Claim-coverage note (r1 F4): the --json-stream shape for this class rides the shared render_results stream machinery (the #170/#198-verified path) — probed surfaces are -J and text.

### DIFF
--- a/riperf3-cli/src/main.rs
+++ b/riperf3-cli/src/main.rs
@@ -349,8 +349,25 @@ fn main() -> std::process::ExitCode {
                         // #267: the lib emits the populated ctrl-closed doc
                         // (bare end{}) before returning this class.
                         | riperf3::RiperfError::ControlSocketClosed
+                        // #374: the client's failed results read — the lib
+                        // emitted the doc with the dangling IERECVRESULTS
+                        // value before returning.
+                        | riperf3::RiperfError::RecvResultsFailed
                 )
             });
+            // #374: GT marks IERECVRESULTS perr, so its errexit line carries
+            // the strerror tail — at the wedge/EOF windows that is the #248
+            // dangling `: ` (GT appends a STALE errno's strerror, live
+            // "Transport endpoint is not connected"; recorded deviation —
+            // the errno-0 form, the #330 server-line precedent).
+            let perr_tail = if e
+                .downcast_ref::<riperf3::RiperfError>()
+                .is_some_and(|le| matches!(le, riperf3::RiperfError::RecvResultsFailed))
+            {
+                ": "
+            } else {
+                ""
+            };
             // --json-stream wins over -J when combined: iperf3's
             // --json-stream gates iperf_json_finish into stream events
             // (review r1 f2, live-verified).
@@ -369,7 +386,7 @@ fn main() -> std::process::ExitCode {
                 // #348: GT's iperf_errexit stamps this line like iperf_err
                 // (iperf_error.c:100-127).
                 let line = format!(
-                    "{}riperf3: error - {e}",
+                    "{}riperf3: error - {e}{perr_tail}",
                     ts_format
                         .as_deref()
                         .map(riperf3::render_timestamp_prefix)
--- a/riperf3-cli/tests/wire_shape.rs
+++ b/riperf3-cli/tests/wire_shape.rs
@@ -4066,3 +4066,196 @@ fn running_phase_err_at_test_start_still_tears_down_streams() {
 fn running_phase_err_mid_running_still_tears_down_streams() {
     running_phase_err_case(true);
 }
+
+// ---------------------------------------------------------------------------
+// #374: the CLIENT's results read gets GT's Nread bounds + the shared
+// Nread_json warning surface (get_results is role-agnostic in GT,
+// iperf_api.c:3036-3080) and the IERECVRESULTS errexit. GT live-probes
+// (3.21, all windows): the state-byte WAITS are UNBOUNDED in GT (silent
+// post-accept, post-TestEnd, and pre-DisplayResults wedges all exceed
+// 45 s) — so riperf3's unbounded recv_state waits are already faithful
+// and stay; only the in-message exchange reads bound. Wedge-hold → exit 1
+// at ~11 s (1 s test + the 10 s idle bound; GT adds its ~10 s close-drain,
+// the #354-recorded residual riperf3 deliberately lacks), the size-read
+// warning, and `error - unable to receive results: ` — the #248 dangling
+// errno-0 form (GT appends a STALE errno's strerror on the EOF path,
+// "Transport endpoint is not connected" live — the #330-recorded
+// deviation, not mirrored). Text prints NO summary. -J: the doc carries
+// the dangling error value and GT's bare `end: {}`.
+// ---------------------------------------------------------------------------
+
+/// How the mock server wedges the results slot after reading the
+/// client's blob.
+#[derive(Clone, Copy)]
+enum ExchangeWedge {
+    /// Send nothing and HOLD the sockets — the size read must self-bound.
+    Hold,
+    /// Send a 500-byte promise + 5 bytes, then HOLD — the blob read must
+    /// self-bound and warn with both counts.
+    PartialBlob,
+    /// Close where the results were due — the fast EOF form.
+    Eof,
+}
+
+/// Mock SERVER for the #374 client pins: full protocol through the
+/// client's TestEnd, sends ExchangeResults, reads the client's results
+/// blob, then wedges per the variant.
+fn exchange_wedge_mock(listener: std::net::TcpListener, wedge: ExchangeWedge) {
+    use std::io::Read;
+    let (mut ctrl, _) = listener.accept().expect("ctrl accept");
+    read_exact(&mut ctrl, 37); // cookie
+    ctrl.write_all(&[9u8]).unwrap(); // ParamExchange
+    read_json_blob(&mut ctrl); // the client's params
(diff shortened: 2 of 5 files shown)


therealevanhenry:
## r1 round record (cold, worktree, @ff98eaa)

**r1: REQUEST-CHANGES — F1 (blocker) fixed @f598dc2; F2/F3 fixed; F4 recorded in the body.**

- **F1 (MAJOR, fixed):** the SERVER_ERROR relay payload read (`read_server_error_payload`) was a bare unbounded `read_exact` — in the issue's plural "exchange reads" scope, and live-proven WORSE than the results-read window: it races no interrupt arm, so the park was signal-immune (SIGTERM did not free it; GT bounds these Nreads and exits a hold at ~30 s with a garbage-errno line from the uninitialized short-read buffer). Fixed on the house Nread budgets with the deterministic None-fallback as the recorded deviation; pin red-first (parked past the 25 s harness bound), M7 added. r1 also corrected my over-narrowed #374 probe-correction claim ("the results read is the client's only in-message read" — factually wrong).
- **F2 (minor, fixed):** `json_read_bounded`'s doc cited the deleted plain `json_read`.
- **F3 (nit, fixed):** the get_results cite (iperf_api.c:2801, called by both roles at :2400/:2404; :3036-3080 is JSON_read).
- **F4 (nit, recorded):** the --json-stream shape rides shared machinery, not probed — claim-coverage note added to the body.

r1 verification: byte-identical server rename confirmed (zero server behavior change); its own GT probes reproduced every pinned surface byte-for-byte; gate 854/854 twice; M6 hand-checked.

## r2 round record (cold, worktree, @f598dc2)

**r2: APPROVE — no code defects; 1 required doc action (stale PR body → updated), 3 nits (F2/F3 applied @9072256 as comment-only records; F4 = the json-stream pin gap, noted).**

r2 verification (independent): every GT source cite verified against 3.21 (net.c:75-76 budgets, Nrecv select mechanics, the payload Nreads, get_results/JSON_read anchors); its own live GT probes reproduced the -J EOF doc (stale-strerror tail, bare end, the -t 1 empty intervals = the recorded #55 drift) and the wedge text surface; the client's ctrl reads independently enumerated (recv_state waits / recv_results / read_server_error_payload — scope complete); the refusal-path blast radius reasoned clean (senders write byte+payload in one burst; a >10 s payload gap breaks GT too — only the rendering differs, recorded); server blast radius clean (pure rename; perr tail unreachable from the server role); M7-equivalent hand-checked; gate **855/855 twice with zero flakes** (both previously-disclosed #349-F4 flakes green under load in all r2 runs); suite-health assessment: the four ~10-11 s pins are timer-idle, degrade to kill+assert (never hang), no cross-test interference — acceptable with the per-PR drift smoke as the watch.

## Merge-head evidence (@9072256 — the r2 F2/F3 comment-only commit)

CI 17/17 green (SHA-keyed watch); mutation table 7/7 red (re-run @f598dc2; the final commit is comment-only); sandbox deploy + compat matrix **52/52 PASS** on this exact head.

Issue #346, opened 2026-07-06
SIGTERM while listening under -J: GT emits a final skeleton doc with the interrupt error key; riperf3 emits nothing


therealevanhenry:
## Divergence

SIGTERM delivered to a listening (idle) server under `-J`:

- iperf3 3.21: emits a final skeleton document carrying `"error": "interrupt - the server has terminated by signal Terminated(15)"` (iperf_err's json sink again), then exits
- riperf3: exits with no `-J` output at all

Text mode is byte-identical between the tools on this path.

## Fix shape

Route the signal-exit path through the same skeleton error-document sink the pre-test failures use, with GT's interrupt sentence. Needs a live GT probe matrix first (SIGINT vs SIGTERM, listening vs mid-test) — mid-test signal surfaces were already done (#225-era); this is specifically the idle-listener case.

therealevanhenry:
GT 3.21 live-probed (SIGTERM at 1s while listening, no client ever):

**-J**: stdout = SKELETON doc — `start`{connected:[], version, system_info} + `intervals`:[] + `end`:{} + `error`: `interrupt - the server has terminated by signal Terminated(15)` (NO `error - ` prefix — the interrupt class), stderr SILENT, **exit 0**.

**Text**: stdout = the listening banner only; stderr `iperf3: interrupt - the server has terminated by signal Terminated(15)`; exit 0.

riperf3 today emits nothing under -J on this path. Fix shape: the listening-phase signal path routes through the same skeleton-doc sink as the pre-test errors (#339 emit_pretest_error_doc precedent) with the interrupt-class key (no prefix); text keeps the existing stderr line. Note GT's observed exit code is 0 (not 128+15) — pin it.

PR #385, opened 2026-07-08
fix: the cookie/params reads race the interrupt watch (#361)

### LINKED ISSUE #361: doc-less -J signal exit remains for a wedged pre-cookie client (the #346 family's residual window)
Found by PR #360's r2 (F2), live-probed both tools. #346 fixed the idle-listening cell, but one doc-less `-J` signal exit remains constructible:

**SIGTERM while a client sits wedged pre-cookie** (connected, silent, never closing): riperf3 exits 0 with EMPTY stdout after the 5 s dump window; GT emits the interrupt skeleton doc immediately.

Mechanism: the pre-cookie reads don't watch the interrupt (the #158 interrupt-blind-by-design window) — the signal only surfaces after the round errors/completes, and that path doesn't route to the #346 accept-site emit.

Same family, zero-width in practice: a signal landing exactly between a round's end and the serve loop's interrupt check (or racing an idle-timeout expiry) also exits doc-less.

Fix shape: either extend the #346 emit to the signal-exit path when NO doc was emitted this round (needs a doc-emitted flag — the double-emit race from PR #360's history says gate carefully), or bound the pre-cookie reads with an interrupt-aware select like the #338 setup wait. Record-deviation is also defensible for the wedged-client sub-cell given the 5 s dump window bounds it.

therealevanhenry:
Closes #361.

## The defect

SIGTERM while a client sits wedged pre-cookie (connected, silent, never closing): riperf3 exited 0 with EMPTY `-J` stdout after the CLI's 5 s dump wall — the cookie/params reads were interrupt-blind, so the signal never surfaced inside the round, and the #346 accept-site emit never ran. GT exits IMMEDIATELY from any phase (its sigend longjmps out of the blocking read): live-probed, the #346 interrupt skeleton at dt 0.00, exit 0 (`-J`: skeleton doc with `"interrupt - the server has terminated by signal Terminated(15)"`; text: the stderr line).

## The fix

`negotiate_test` (the cookie + params reads) races `wait_interrupt` in a select — the interrupt arm emits the exact #346 skeleton (`emit_pretest_error_doc`) and ends the round `Ok(None)`; the serve loop's interrupt check exits 0. Cancel-safety: the round is over either way, the ctrl drops.

Two tests carried the old blind-window semantics and were reworked in the same commit:

- **The #158 second-signal wedge** moved one phase later: params sent, no data connections — the CREATE_STREAMS setup wait has no interrupt arm and a 120 s default rcv_timeout, so the first signal's 5 s dump window still holds deterministically for the second-signal hard-exit test. (If a future issue makes the setup waits interrupt-aware — the same GT-parity argument applies there — the test moves again or gets a synthetic wedge.)
- **The #360 refused-round single-doc test**: its scenario (signal parked in recv_params, then violating params) now ends at the signal — GT never reaches the refusal. The test asserts the GT shape (ONE doc, the interrupt key) and keeps its no-double-doc whole-stdout guard.

Scope note: the setup waits (TCP accept select, both UDP selects) remain interrupt-blind — bounded by rcv_timeout, but a signal there still rides the dump wall. Same family, one phase later; not in #361's probed scope — noted on the issue as a possible follow-up.

## Verification

**Pins, red-first** (both rode the 5 s wall — the `elapsed < 3 s` assert is the red):

- `sigterm_on_wedged_precookie_client_emits_skeleton_in_json` — the skeleton emits immediately, exit 0, silent stderr.
- `sigterm_on_wedged_precookie_client_prints_line_in_text` — GT's stderr line, exit 0, immediate.

**Mutation table:** M1 interrupt arm removed (the blind window back) → red; M2 skeleton emit dropped → red; M3 the arm errors instead of the signal-normal end → red (text pin's exit-0/line assert).

**Gate:** fmt / clippy `-D warnings` clean; workspace **876 passed / 0 failed**. One full-run flake (`end_loop_unknown_byte_takes_gt_iemessage_shape_in_json`, the #349-F4 load family — green 3/3 isolated, green on the full rerun); disclosed, no change.

**Rounds:** r1 APPROVE (F1 stale comment + F3 json-stream pin + F4 inert flag applied; F2 → #386, the pre-existing refusal-round-end divergence). r2 APPROVE (cold, 18/18 GT cells byte-parallel; the two one-line asserts applied). Full round records in the PR comments. Merge-head evidence: CI 17/17 on every head; compat 52/52; mutants 3/3.

### DIFF
--- a/riperf3-cli/tests/pidfile.rs
+++ b/riperf3-cli/tests/pidfile.rs
@@ -139,10 +139,13 @@ fn pidfile_unlinked_after_one_off_run() {
 /// #158/#210: a SECOND signal exits immediately, dying BY the signal. The
 /// first signal now takes the graceful sigend path (#210: dump + terminate
 /// the peer), so the old blasting-UDP wedge converges too fast to race.
-/// Instead a half-open control connection wedges the server in the param
-/// exchange — a phase with no interrupt-aware await — so the first signal's
-/// bounded dump window (5 s) holds deterministically while the second signal
-/// hits the pre-armed raw handler.
+/// The deterministic wedge is a phase with no interrupt-aware await: the
+/// cookie/param reads served until #361 made them interrupt-aware (GT
+/// exits immediately there), so the wedge now parks in the CREATE_STREAMS
+/// setup wait — params sent, no data connections; its select watches the
+/// ctrl and the rcv_timeout (120 s default) but NOT the interrupt — so the
+/// first signal's bounded dump window (5 s) holds deterministically while
+/// the second signal hits the pre-armed raw handler.
 #[cfg(unix)]
 #[test]
 fn second_signal_during_teardown_exits_immediately() {
@@ -168,10 +171,25 @@ fn second_signal_during_teardown_exits_immediately() {
         "server pidfile written",
     );
 
-    // The wedge: connect to the control port and send nothing — the server
-    // sits in the cookie/param read, which no interrupt arm covers.
-    let _wedge = std::net::TcpStream::connect(("127.0.0.1", port)).expect("wedge connect");
-    std::thread::sleep(Duration::from_millis(300)); // let the accept land
+    // The wedge (#361 moved it one phase later): complete the cookie/param
+    // exchange, then connect NO data streams — the server parks in the
+    // CREATE_STREAMS setup wait, which no interrupt arm covers.
+    let _wedge = {
+        use std::io::{Read as _, Write as _};
+        let mut w = std::net::TcpStream::connect(("127.0.0.1", port)).expect("wedge connect");
+        w.write_all(&[b'x'; 37]).expect("cookie");
+        let mut b = [0u8; 1];
+        w.read_exact(&mut b).expect("ParamExchange state");
+        assert_eq!(b[0], 9);
+        let params = br#"{"tcp":true}"#;
+        w.write_all(&(params.len() as u32).to_be_bytes())
+            .expect("len");
+        w.write_all(params).expect("params");
+        w.read_exact(&mut b).expect("CreateStreams state");
+        assert_eq!(b[0], 10);
+        w
+    };
+    std::thread::sleep(Duration::from_millis(300)); // let the setup wait park
 
     let pid = server.0.id() as i32;
     // SAFETY: plain kill(2) on our own child.
--- a/riperf3-cli/tests/signal_terminate.rs
+++ b/riperf3-cli/tests/signal_terminate.rs
@@ -971,15 +971,16 @@ fn server_sigterm_mid_test_emits_exactly_one_doc() {
     );
 }
 
-/// #346 (PR #360 r1 F1): a SIGTERM landing DURING a refused round must not
-/// append the idle-interrupt skeleton after the refusal doc — the refusal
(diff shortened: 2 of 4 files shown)


therealevanhenry:
## r1 round record (cold, worktree, @3d63927)

**r1: APPROVE — 4 findings: F1 (stale comment) + F3 (json-stream pin) + F4 (inert flag) applied @850a8f6; F2 → #386.**

- **F1 (minor, applied):** accept_control's doc still claimed the cookie/param reads stay interrupt-blind by design — both halves false post-#361; now points at the remaining setup-wait window.
- **F2 (minor) → #386 filed:** pre-existing refusal-round-end divergence surfaced by r1's hammer probe — riperf3's refused round completes immediately (two docs under the signal race: refusal + next-round skeleton, 30/30) where GT parks doc-less until client EOF and the sigend abandons the unprinted refusal doc (skeleton only, 10/10). Verified pre-existing via the M1 mutant.
- **F3 (nit, applied):** the --json-stream wedged pin (symmetric with the #346 idle cell — probed identical on both tools).
- **F4 (nit, applied):** the reworked #360 test dropped its inert --server-max-duration flag.

r1 verification: 6/6 GT cells probed (pre-cookie AND mid-params × -J/text/json-stream — the select covers both reads); cancel-safety proven structurally (both refusal helpers write-then-print with no await between print and return — no strandable printed doc); the watch borrow/clone pattern verified against tokio source (no missed-signal window); the #158 rework's determinism verified (all three setup selects interrupt-arm-free; auth never reads); M1 hand-checked; 50 green signal-test executions incl. 2-core pinned.

## r2 round record (cold, worktree, @850a8f6)

**r2: APPROVE — no blockers; F1/F2 (one-line asserts) applied @31a7c56; F3/F4 noted; F5 = the recorded setup-wait residual.**

- **F1 (minor, applied):** the json-stream pin's bare-end-event data assert (the #346 twin's).
- **F2 (nit, applied):** the -J pin's skeleton `connected: []` assert (GT shows it even with the client attached).
- **F3 (nit, noted):** drive_sigterm_wedged duplicates the listening driver; refactor deferred (churn > value mid-batch).
- **F4 (nit):** the stale 876 count in the body prose — corrected below: the final suite is **878** (877 + the r2 asserts ride existing tests; 877 was post-F3).
- **F5 (observation):** the zero-width same-poll race into the setup wait — correctly recorded as the #361 scope comment's residual.

r2 verification (independent): 18/18 GT cells byte-parallel (three wedge positions × three sinks, plus the #346 idle cell undisturbed); the #386 race sweep re-probed (skeleton-only convergence in-cell); tokio watch version semantics verified in source (clone preserves seen-version; borrow never marks); the refusal helpers' no-double-emit structure re-derived; all three setup selects confirmed interrupt-arm-free; 50 green pinned/unpinned signal-test runs; M1 hand-checked; gate 877/877 zero flakes.

## Merge-head evidence (@31a7c56 — the r2 test-only asserts)

CI 17/17 (SHA-keyed watch); mutation table **3/3 red** (@3d63927; subsequent commits comment/test-only); sandbox deploy + compat **52/52 PASS** (@850a8f6; the final commit adds two test asserts — no production delta); one full-suite load flake per run at most (the disclosed #349-F4 family), green isolated and on rerun each time.

Issue #343, opened 2026-07-06
results blob with trailing garbage: GT accepts (cJSON), riperf3 rejects (serde) — mirror vs record


therealevanhenry:
## Divergence (decision needed)

GT parses the results blob with `cJSON_Parse` (`require_null_terminated=0`), so **valid JSON followed by trailing garbage inside the promised length parses fine** and the exchange succeeds (live-confirmed: GT sends its results and displays normally).

riperf3's `serde_json::from_slice` rejects trailing bytes, so the same blob fails the exchange with IERECVRESULTS.

Practical impact: a sloppy third-party iperf3 implementation that over-declares its length works against iperf3 and fails against riperf3.

## Options

1. Mirror GT: parse leniently (e.g. `serde_json::Deserializer::from_slice(...).into_iter()` first-value, or trim at the first complete value).
2. Record as a deviation: strict parsing is arguably the cleaner behavior; GT's laxity is a cJSON artifact, not a designed protocol feature.

Leaning (2) per the project posture of not mirroring GT accidents — but it needs an explicit call and a deviation record either way.

therealevanhenry:
GT 3.21 live-probed (probe343.py): a results blob of VALID JSON + trailing garbage inside the declared length is ACCEPTED — GT completes the exchange, sends its own results, full clean round, rc=0, empty stderr. Same for the PARAMS blob (garbage-suffixed params → CREATE_STREAMS follows). cJSON_Parse consumes the first JSON value and ignores the remainder; GT relies on that everywhere it reads a blob.

**DECISION: MIRROR.** This is deliberate library leniency GT depends on, not bug noise (no miscount, no nonsense surface) — the don't-mirror rule covers bugs; wire-compat leniency is the faithful posture (the #328 atoi-fidelity precedent). Fix: prefix-lenient parse at BOTH blob read sites (recv_params + recv_results\*) — serde_json StreamDeserializer first-value semantics; garbage from byte 0 still rejects in both tools. Pin both accept-with-suffix cells + keep the reject-from-byte-0 pins.

PR #400, opened 2026-07-08
fix: mirror cJSON's BOM skip and bare-scalar params root; record 4 leniency deviations (#367)

### LINKED ISSUE #367: five residual cJSON-leniency cells beyond the #343 mirror (BOM, scalar roots, adjacent garbage, deep nesting, control chars)
Found by PR #366's r1 (F2), all live-probed on GT 3.21 and identical on main (pre-existing, NOT regressions of the #343 mirror). Five residual cJSON-leniency cells where GT accepts a blob riperf3 rejects:

1. **UTF-8 BOM + valid JSON** (both blob sites) — cJSON `skip_utf8_bom`; serde rejects. (The #343 record's "garbage from byte 0 still rejects in both tools" is technically wrong for this one cell.)
2. **Bare-scalar params root** (`42`) — GT's `get_parameters` is all-optional key-lookup and proceeds to CREATE_STREAMS with defaults; riperf3's `from_value::<TestParams>` requires a map.
3. **Scalar with adjacent garbage** (`42GARBAGE`) — cJSON stops at the first non-value byte; serde's StreamDeserializer rejects non-self-delimiting values not followed by whitespace/EOF.
4. **Nesting depth 128–999** — serde's recursion limit is 128; `CJSON_NESTING_LIMIT` is 1000.
5. **Raw control characters inside strings** (`\x01`) — cJSON copies them; serde rejects.

None are reachable from a real iperf3 peer's encoder (its cJSON printer emits none of these shapes), so the practical interop exposure is hostile/nonconforming peers only. The #343 helper's scope comment records the boundary (self-delimiting first values, whitespace-separated tails).

Decide per-cell: mirror (e.g. BOM skip is one line), or record-deviation. A client-side e2e pin for the shared helper's third site is also pending a mock-server harness (r1 F4).

therealevanhenry:
Closes #367.

Six residual cJSON-leniency cells beyond the #343 first-value mirror, where GT's cJSON accepts a blob riperf3's serde rejects. Resolved per-cell: **mirror the two cheap/safe cells, record the four that would need a bespoke lenient parser** (none are reachable from a real iperf3 encoder — hostile/nonconforming peers only).

## Mirrored

- **Cell 1 — UTF-8 BOM.** cJSON runs `skip_utf8_bom` at every parse entry (cjson.c:1074-1088, called :1131), so a BOM ahead of valid JSON parses on every blob. `json_first_value` now strips a leading `EF BB BF`. This also corrects the #343 record's "garbage from byte 0 still errors in both tools" — untrue for this one cell.
- **Cell 2 — bare-scalar params root.** GT's `get_parameters` (iperf_api.c:2533+) is a sequence of typed `cJSON_GetObjectItem` lookups; on a non-object root every lookup misses → no field set → GT proceeds to CREATE_STREAMS with defaults. `params_from_value` now maps any non-object root (scalar/array/string/bool/null) to `TestParams::default()`. Scoped to params — results keep their required-field validation.

## Recorded deviations (locked by test)

serde is a conforming JSON parser; mirroring these means replacing it with a cJSON-clone, disproportionate for shapes no real iperf3 peer emits:

- **Cell 3** — scalar with adjacent non-whitespace garbage (`42GARBAGE`): cJSON stops at the first non-value byte; serde requires a non-self-delimiting first value to be whitespace/EOF-terminated.
- **Cell 4** — nesting past serde's 128-deep recursion limit (cJSON's `CJSON_NESTING_LIMIT` is 1000).
- **Cell 5** — raw control chars inside strings: cJSON copies them; serde rejects unescaped control bytes.
- **Cell 6** — cJSON's lenient strtod grammar (`01`, `1.`): serde rejects.

`json_first_value_deviations_stay_strict` locks all four — a future serde change that silently began accepting any of them trips there.

## GT verification

All GT claims source-verified this session: cjson.c `skip_utf8_bom`; iperf_api.c `get_parameters` typed key-lookup. The six cells' GT-accept behavior was live-probed on GT 3.21 when the issue was filed (PR #366 r1 F2 + r2).

## Tests (red-first) + mutation table

- `json_first_value_skips_utf8_bom` — red pre-fix (serde rejects the BOM), green after.
- `params_from_value_defaults_on_non_object_root` — red pre-fix (from_value errors on a scalar), green after; the object-root arm still parses its fields.
- `json_first_value_deviations_stay_strict` — green (locks cells 3-6).
- 6/6 mutation table killed (BOM strip removed / mangled bytes; is_object gate inverted / fallback dropped / object arm defaulted; a leniency-creep mutant that the deviation lock catches). Tree restored clean.

## Disclosure

One `mss_blksize` load flake (UDP MSS-after-cookie timing, unrelated to JSON parsing) in one of three full-workspace runs; green isolated and in the other two.

### DIFF
--- a/riperf3/src/protocol.rs
+++ b/riperf3/src/protocol.rs
@@ -298,18 +298,29 @@ pub async fn json_write(stream: &mut TcpStream, value: &serde_json::Value) -> Re
 // Test parameters — exchanged as JSON during ParamExchange
 // ---------------------------------------------------------------------------
 
+/// UTF-8 BOM (#367 cell 1): cJSON's parse entry runs `skip_utf8_bom` before
+/// `parse_value` (cjson.c:1074-1088, :1131), so a leading BOM ahead of valid
+/// JSON parses on every blob. serde has no such skip; strip it here to match.
+const UTF8_BOM: &[u8] = b"\xEF\xBB\xBF";
+
 /// #343: GT parses every length-prefixed blob with
 /// cJSON_Parse(require_null_terminated=0) — the first JSON value wins and
 /// trailing bytes inside the declared length are ignored. Deliberate wire
 /// leniency GT depends on (a peer that over-declares its length
-/// interoperates), mirrored per the #328 fidelity precedent; garbage from
-/// byte 0 still errors like both tools. SCOPE (r1 F2): the mirror covers
-/// self-delimiting first values with whitespace-separated tails — the
-/// residual cJSON cells (UTF-8 BOM, bare-scalar roots, scalar-adjacent
-/// garbage, nesting past serde's 128, raw control chars in strings) stay
-/// divergent and are tracked in the follow-up issue; none are reachable
-/// from a real iperf3 encoder.
+/// interoperates), mirrored per the #328 fidelity precedent.
+///
+/// SCOPE (#343 r1 F2 / #367): the mirror covers self-delimiting first values
+/// with whitespace-separated tails, plus a leading UTF-8 BOM (#367 cell 1,
+/// now stripped here — cJSON's skip_utf8_bom). The bare-scalar params root
+/// (#367 cell 2) is mirrored in `params_from_value`. Four residual cJSON
+/// cells stay divergent as RECORDED DEVIATIONS — serde is a conforming
+/// parser and mirroring them means a bespoke lenient parser for shapes no
+/// real iperf3 encoder emits: scalar-adjacent garbage (`42GARBAGE`), nesting
+/// past serde's 128-deep limit (cJSON's is 1000), raw control chars in
+/// strings, and cJSON's lenient strtod grammar (`01`, `1.`). Locked by
+/// `json_first_value_deviations_stay_strict`.
 fn json_first_value(buf: &[u8]) -> serde_json::Result<serde_json::Value> {
+    let buf = buf.strip_prefix(UTF8_BOM).unwrap_or(buf);
     let mut stream = serde_json::Deserializer::from_slice(buf).into_iter();
     match stream.next() {
         Some(v) => v,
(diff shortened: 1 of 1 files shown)


therealevanhenry:
**Review round 1 (cold adversarial agent, worktree @8b66f24): APPROVE, no blockers.**

The reviewer compiled a C probe against the bundled cjson.c (mirroring `JSON_read`'s `calloc(strlen+1)` + `cJSON_Parse`) and enumerated every root shape, plus reverted each fix hunk to confirm the pins bite red:

- BOM mirror faithful at every edge: BOM-only (strlen 3) and BOM+space both → GT NULL/ERROR and riperf3 Err (cJSON's `can_access_at_index(buffer,4)` guard means it only skips when there's ≥1 content byte); BOM+content → both proceed. The shared helper applying BOM-skip to results is correct — GT's `get_results` runs the same `cJSON_Parse`/`skip_utf8_bom`; results keep their required-field validation.
- Bare-scalar root faithful for `42`/`true`/`null`/`"hi"`/`[1,2,3]`: cJSON returns a non-NULL item for each (incl. `null` as a valid `cJSON_NULL`), so GT never hits `j==NULL → IERECVPARAMS`; every typed lookup misses → defaults → CREATE_STREAMS. No crash on an array root. Empty blob is caught upstream (both error).
- All four recorded deviations genuinely divergent (cJSON accepts, serde rejects), confirmed by probe + source (`parse_string` copies raw control bytes; `CJSON_NESTING_LIMIT`=1000 vs serde 128; strtod leniency).
- No common-path regression (44 protocol round-trip tests green; normal blobs start with `{` so the BOM strip is a no-op).

Observations, disposition:
- **OBS1**: a genuine 7th divergence — a wrong-typed object field (`{"time":"foo"}`) makes GT warn (`iperf_cJSON_GetObjectItemType mismatch`, iperf_util.c:464) + default the field + run, where riperf3's `from_value` hard-errors. Honestly scoped out of #367. **Filed #401.**
- **OBS2**: the comment said "typed `cJSON_GetObjectItem` lookups"; the actual call is the `iperf_cJSON_GetObjectItemType` wrapper. **Fixed @d968a54** (also references #401).
- **OBS3**: cells 2 and 3 both map to GT "non-object → defaults"; riperf3 mirrors cell 2 and diverges on cell 3 only because serde succeeds on `42` and errors on `42GARBAGE` — a coherent, documented boundary, no action.

therealevanhenry:
**Review round 2 (fresh cold adversarial agent, worktree @d968a54): APPROVE, no blockers.**

Independent re-derivation with a C probe against the exact vendored cjson.c (1.7.15):
- BOM equivalence holds at every edge, including the interior-NUL case (the strlen-vs-length model can't diverge — a NUL never continues a serde first value; `42\0.5` → 42 on both, `[1,\02]` → error on both).
- Double-BOM strips only one (offset-0 guard) on both; interior BOM strips on neither.
- Every non-object root defaults on both (incl. i64-overflow parsing as f64 → default, and empty-object routing through `from_value({})` → all-defaults); none hit GT's `j==NULL` error path; array root doesn't crash.
- All four deviation cells fail for the CLAIMED reason (bystander probes: bare `42`/`"ab"`/100-deep all parse, so cell 3 = adjacent garbage, cell 5 = the control byte, cell 4 = depth not length).
- Red-first re-verified; the deviation lock is a genuine tripwire (green with or without the fix); 413/413 lib tests; results-path BOM is a symmetric fidelity improvement; all source citations accurate.

**F1 (non-blocking)**: a further divergence class — cJSON's `buffer_skip_whitespace` treats every byte `<=0x20` as whitespace (cjson.c:1061), so control bytes in whitespace positions are tolerated where serde rejects them. Distinct from cell 5 (control byte *inside a string*). Pre-existing, not introduced here, same recorded-deviation posture. **Filed #402** to keep the deviation ledger complete.

Two cold rounds complete (r1 + r2, both APPROVE). Ready to merge after deploy + compat 52/52 on the merge head.

Issue #351, opened 2026-07-07
server: missing TEST_RUNNING data-idle watchdog (GT's IENOMSG=144 restart path)


therealevanhenry:
## Gap (found in PR #349 r1)

GT's server has a data-idle restart path riperf3 lacks entirely: when no messages/data arrive for `NO_MSG_RCVD_TIMEOUT` while a test is active, the server errors with **IENOMSG(144)** ("idle timeout while waiting for data"), relays it via cleanup_server (iperf_server_api.c:720-730), and restarts to serve the next client.

riperf3 has no IENOMSG code path and no equivalent data-idle watchdog on a running test. (The #145/#224 idle handling covers different windows.)

## Relation to open work

#338's fix adds GT's 120 s no-progress bound to the CREATE_STREAMS accept window (rcv_timeout, iperf_server_api.c:663-678) — that is the pre-test sibling of this same watchdog family. This issue tracks the TEST_RUNNING-phase half: detection, the IENOMSG surface (text + -J + wire relay), and keep-serving semantics.

therealevanhenry:
GT 3.21 live-probed (probe351.py: --rcv-timeout 3000, mock reaches TEST_RUNNING, sends 4 KiB, then goes SILENT holding both sockets):

- **Wire**: SERVER_ERROR relay `fe 00000090 00000000` (IENOMSG=144, errno 0) at dt=3.0s — deterministic on the timeout.
- **Text**: stderr `iperf3: error - idle timeout for receiving data`; stdout keeps ticking ZERO-byte interval rows while idle until the watchdog fires; NO summary block; rc=0 (keep-serving class).
- **-J**: silent stderr; ONE doc with the accumulated intervals (0-byte rows PRESENT) + bare `end:{}` + error key `error - idle timeout for receiving data` (WITH the `error - ` prefix); rc=0.

Implementation notes: riperf3 already has `--rcv-timeout <ms>` but wires it CLIENT-side only (SO_RCVTIMEO on data streams, client.rs:1167). GT's SERVER uses the same knob (default 120000 ms, DEFAULT_NO_MSG_RCVD_TIMEOUT iperf_api.h:70) for this watchdog — the fix adds the server-side idle timer to the TEST_RUNNING loop: reset on any data/ctrl progress, on expiry take the IENOMSG surface above (relay + bare-end doc with accumulated intervals + stderr line + keep serving). The #338 fix's 120 s CREATE_STREAMS bound is the pre-test sibling using the same knob.

[...]

Issue #230, opened 2026-06-11
max-duration: missing iperf3's upfront requested-duration check at param exchange


therealevanhenry:
From #224's ground-truth session (iperf 3.21, pinned interop build, 2026-06-11): iperf3's `--server-max-duration` has an **upfront check at param exchange** that riperf3 lacks entirely — riperf3 only implements the mid-test timer arm.

Observed iperf3 behavior (both shapes verified live):

- Client requests `-t 6` against `--server-max-duration 2` → rejected **before the test starts**: `SERVER_ERROR` relay, client stderr `iperf3: SERVER ERROR - client's requested duration exceeds the server's maximum permitted limit` + the adopted errexit line, exit 1. Server prints the same message to its stderr, one-off exits 0.
- A `-n 500G` run (duration unknowable upfront) is **also rejected upfront** — iperf3 treats unbounded-duration requests as exceeding the limit rather than letting the wall-clock timer catch them.

So with #224 fixed (the timer arm now relays `SERVER_ERROR` + `IESERVERTESTDURATIONEXPIRED` faithfully), the remaining divergence is that riperf3 *runs* tests iperf3 would refuse at param exchange, and the timer only fires after the fact.

Fix shape: at param exchange, when `server_max_duration` is set and the client's requested duration (or an unbounded `-n`/`-k`/`-t 0` request) exceeds it, send the `SERVER_ERROR` relay with iperf3's errno for `client's requested duration exceeds the server's maximum permitted limit` (find the exact IE code in iperf_api.h — it is NOT IESERVERTESTDURATIONEXPIRED=160, whose strerror is the different "server test duration expired" string), print the same line to the server's sink, and never start the test. `self_terminate.rs::max_duration_timer_relays_expired` deliberately uses `-n` so it pins the TIMER arm and survives this change — but it asserts today's pre-upfront-check behavior on that path, so this fix must revisit that test (the `-n` run will then be rejected upfront; the timer arm needs a different trigger, e.g. a slow `-t`-within-limit run that overruns on wall clock).

Blast radius: param-exchange behavior change → own PR, post-#224.

therealevanhenry:
[...]

Issue #224, opened 2026-06-11
Server self-terminate (bitrate-limit/max-duration) still runs the full results exchange after SERVER_TERMINATE


therealevanhenry:
From #223's round-1 review (n8): the `server_terminated` path (--server-bitrate-limit / --server-max-duration sending SERVER_TERMINATE) still proceeds to the ExchangeResults/DisplayResults/IperfDone handshake after telling the peer to stop — the peer never exchanges, so the server dies on PeerDisconnected instead of dumping cleanly. #223's skip-gate covers only the client_terminated/interrupted paths. Also `server_results` is dead work on all terminate paths. Fix shape: fold `server_terminated` into the same skip-gate + local dump.

PR #403, opened 2026-07-08
fix: rate-breach and duration-watchdog -J docs render bare end{} like GT (#368)

### LINKED ISSUE #368: self-terminate -J docs render a populated end where GT keeps bare end{} (rate-breach + likely duration-expiry; the #351 idle arm already matches)
Found while probing GT for #351 (live, GT 3.21): on the **runtime `--server-bitrate-limit` breach**, GT's `-J` document renders the accumulated intervals over a **bare `end:{}`** (its kill path — cleanup_server + return -1 — never runs end processing; probed: `end` keys `[]`, prefixed error key, intervals as accumulated). riperf3's #224-era implementation renders the **fully populated finalize end** (streams/sums/cpu/congestion) on the same cell.

The #351 idle-watchdog arm already takes the bare-end shape (`ctx.bare_end`, pinned); the rate-breach sibling needs the same one-line flag at its arm — plus a check of the duration-expiry (`--server-max-duration`, IESERVERTESTDURATIONEXPIRED 160) sibling, which likely shares the class (probe GT first).

Text-mode note: riperf3 already skips the summary block on these paths (#224); this issue is the `-J` end-block shape only. Existing #224-era pins asserting the populated end will need realigning with the GT probe.

therealevanhenry:
Closes #368.

On a runtime **`--server-bitrate-limit` breach** (and its **duration-watchdog** sibling), GT's `-J` document renders the accumulated intervals over a **bare `end: {}`** — the kill path frees the streams and returns without running end processing. riperf3's #224-era code rendered the fully populated finalize end (streams/sums/cpu/congestion) on these arms. The idle-watchdog arm (#351) already sets `ctx.bare_end`; this extends the same one-line flag to the two siblings.

## GT probes (3.21, live)

| cell | GT `-J` end | riperf3 pre-fix |
|---|---|---|
| rate breach (`--server-bitrate-limit 1K`, `-t 5`) | `end` keys `[]`, error key, exit 0 | populated: `streams/sum_sent/sum_received/cpu_utilization_percent/receiver_tcp_congestion` |
| duration watchdog (wedged client past `dur+omit+40s`) | `end` keys `[]` | populated |

GT source: rate path `cleanup_server` + `return -1` (iperf_server_api.c:624-646); duration path `server_timer_proc` frees streams + sends SERVER_ERROR (iperf_server_api.c:318-337) — neither reaches the TEST_END → DISPLAY_RESULTS end processing that populates the doc.

Note on the duration probe: with a wedged client, GT's error key clobbers to a stale `select failed: Bad file descriptor` (the same `i_errno`-global nondeterminism recorded for #349/#350) — but the end **shape** is bare, which is all #368 concerns. The error-key content stays riperf3's intended `SELF_TERM_DURATION_MSG` (pinned since #224).

## Fix

`ctx.bare_end = true` at the rate-breach arm and the duration-watchdog arm in `await_test_end` — identical to the idle arm. The flag flows to `input.bare_end` in the server's report builder; intervals and the start block are untouched (only the end block is suppressed).

## Tests + mutation table

- `bitrate_limit_json_server_doc_bare_end_and_prefixed_error` (renamed from `..._carries_prefixed_error`, whose "intervals and all" comment was the stale claim) — red-first: asserts bare `end: {}` + the prefixed error key + intervals/start survive. Red pre-fix (populated end), green after.
- `get_server_output`'s server-own-doc read gains an inverse guard: a NORMAL run keeps a populated end (the flag must not fire on the happy path) — this was an untested property (mutation M3 initially survived without it).
- Mutation table 3/3: M1 rate-arm flag dropped, M2 consumer wiring forced false (both red the rate pin), M3 consumer forced true (reds the inverse guard). Tree restored clean.

**Duration arm — not CI-pinnable**: the watchdog fires at `duration + omit + 40s` and the 40s grace can't be shortened (same as the #230 watchdog-timing behavior), so a ~41s wedged-client test doesn't belong in CI. It's a one-line flag identical to the rate/idle siblings, GT-probe-confirmed bare, verified by inspection.

Full workspace green.

### DIFF
--- a/riperf3-cli/tests/get_server_output.rs
+++ b/riperf3-cli/tests/get_server_output.rs
@@ -178,6 +178,13 @@ fn json_client_gets_server_output_json() {
         own_n > 0,
         "the -J server's own doc lost its intervals (double-build class): {server_own}"
     );
+    // #368 inverse guard: a NORMAL (non-self-terminated) server run keeps the
+    // POPULATED finalize end — the bare-end flag must not fire on the happy
+    // path (only the rate/duration/idle kill arms set it).
+    assert!(
+        sv["end"].as_object().is_some_and(|m| !m.is_empty()),
+        "a normal server run's end must stay populated: {server_own}"
+    );
     assert_eq!(
         Some(own_n),
         sj["intervals"].as_array().map(Vec::len),
--- a/riperf3-cli/tests/self_terminate.rs
+++ b/riperf3-cli/tests/self_terminate.rs
@@ -145,11 +145,15 @@ fn bitrate_limit_json_client_single_doc_with_relayed_error() {
     );
 }
 
-/// -J server: the full document still emits (intervals and all) with the
-/// error key carrying iperf_err's `error - ` prefix wart, stderr empty,
-/// exit 0 — live-verified against iperf 3.21.
+/// -J server: the accumulated intervals emit over a BARE `end: {}` (#368) —
+/// GT's rate-breach kill path (cleanup_server + return -1,
+/// iperf_server_api.c:624-646) never runs end processing, so the reader's
+/// document finishes with `end` as an empty object; the error key carries
+/// iperf_err's `error - ` prefix wart, stderr empty, exit 0. Live-verified
+/// against iperf 3.21 (`--server-bitrate-limit 1K` + `-t 5`: GT `end` keys
+/// `[]`, riperf3 pre-#368 rendered the full finalize end).
 #[test]
-fn bitrate_limit_json_server_doc_carries_prefixed_error() {
+fn bitrate_limit_json_server_doc_bare_end_and_prefixed_error() {
     let ps = common::free_port().to_string();
     let server = spawn_server(&["--server-bitrate-limit", "1K", "-J"], &ps);
     std::thread::sleep(Duration::from_millis(300));
@@ -172,6 +176,19 @@ fn bitrate_limit_json_server_doc_carries_prefixed_error() {
         doc["error"].as_str(),
         Some(format!("error - {BITRATE_MSG}").as_str()),
         "iperf_err's in-doc prefix wart, mirrored exactly: {doc}"
+    );
+    // #368: the kill path skips end processing — bare `end: {}`, NOT the
+    // populated finalize end (streams/sums/cpu/congestion).
+    assert_eq!(
+        doc["end"].as_object().map(serde_json::Map::len),
+        Some(0),
+        "GT's rate-breach end is bare {{}}: {doc}"
+    );
+    // The accumulated interval rows still ride the doc (the start block too);
+    // only the end block is suppressed.
+    assert!(
+        doc["intervals"].is_array() && doc["start"].as_object().is_some_and(|m| !m.is_empty()),
+        "intervals + start survive the bare-end path: {doc}"
     );
 }
 
--- a/riperf3/src/server.rs
+++ b/riperf3/src/server.rs
@@ -2075,6 +2075,14 @@ impl Server {
                                 // SERVER_TERMINATE (iperf 3.21 GT).
                                 protocol::send_server_error(&mut ctx.ctrl, 27).await?;
                                 ctx.server_error = Some(SELF_TERM_RATE_MSG);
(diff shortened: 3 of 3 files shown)


therealevanhenry:
**Review round 1 (cold adversarial agent, worktree @75edfc7): APPROVE, no blockers.**

The reviewer built the worktree and probed GT vs riperf3 on the rate cell (`--server-bitrate-limit 1K -J`): both render bare `end: {}` with the `error - ` prefix wart and preserved start/intervals — shape/key-order match. It also cleared the two angles that could have been blockers:

- **Capture-drop (the most-likely-bug angle): NOT a regression.** `exchange_results_phase` is gated on `ctx.server_error.is_none()` (server.rs:2446), so on breach the server never enters ExchangeResults → never sends `server_output` — matching GT (probed: neither attaches it). `ctx.capture` is `None` in `-J` mode anyway; the `.filter(|_| !ctx.bare_end)` only affects the server's own text console, where the pre-fix spurious final summary is now suppressed — strictly MORE faithful (GT prints none on breach). Text summary was already `server_error.is_none()`-gated, so text output is unchanged.
- **Duration-arm source reasoning: sound.** `server_timer_proc` sets `done=1` + closes `ctrl_sck` but not `IPERF_DONE`; the next `select()` on the stale fd returns EBADF → `cleanup_server` + `return -1` (iperf_server_api.c:681-684), the SAME error-exit as the rate breach, never reaching `iperf_json_finish` → bare end. No partial dump. The "select-fail clobber" nondeterminism note is accurate.

Both pins bite (reverting the fix reds the rate pin with the full populated payload; forcing `input.bare_end=true` reds the inverse guard). Rate pin is not flaky (breach at the first 1 Hz tick ~1s, far inside `-t 5`). Rename left no dangling references.

Findings, disposition:
- **F1 (out of scope)**: the CLIENT's own `-J` doc is still divergent — a mid-run SERVER_ERROR relay renders a populated end where GT is bare (`on_server_error_relay`'s `bare_end = !ctx.stage.started()` stays false mid-run). Pre-existing, client-side; #368 is server-side. **Filed #404** (client twin).
- **F2 (accepted)**: the duration arm has no end-shape test (a -J bare-end pin would need a second ~45s watchdog test — a deliberate cost trade). The flag is one line identical to the pinned rate arm, on the pinned consumer plumbing, GT-probe-confirmed; inspection-only is defensible. Left as-is.

No code change (both findings non-blocking); head stays @75edfc7.

therealevanhenry:
**Review round 2 (fresh cold adversarial agent, worktree @75edfc7): APPROVE, no blockers.**

Independent re-derivation of all seven angles, with GT probes:
- **`-J` and `--json-stream` are consistent** and both match GT byte-for-byte on a rate breach: `-J` end `{}`; `--json-stream` tail `{"event":"error",...}` then `{"event":"end","data":{}}` — identical to GT. Structural: both modes render `report.end`, whose manual Serialize short-circuits to `{}` when bare (json_report.rs:724), and both build through the same `input.bare_end` consumer.
- Arms are mutually-exclusive `tokio::select!` branches; `bare_end` inits false and is only ever set true (idempotent) — no double-set/ordering issue.
- Inverse guard reads the server's OWN `-J` stdout doc (not the client/attachment); reds when `bare_end` is forced true.
- No wire/exit-code/teardown change (server exit 0, client exit 1, SERVER_ERROR(27) unchanged; bare_end applied after send_server_error, render-only).
- Both GT comment claims re-verified against source + probe (rate `return -1` before end processing; duration `server_timer_proc` → select EBADF → cleanup_server + return -1).

Notable non-blocking observations:
- **NF2 (bonus)**: in TEXT mode `--get-server-output`, `ctx.capture` IS `Some`, so the `!ctx.bare_end` filter now drops the capture on a breach — and this MATCHES GT (GT text server prints live interval lines but no final summary on breach; pre-PR riperf3 rendered a summary block). The PR silently corrects a text-mode divergence too.
- **NF1**: confirms #404 (client-side twin) is a real, correctly-deferred divergence — orthogonal (separate role, no internally-inconsistent doc), pre-existing.

Two cold rounds complete (r1 + r2, both APPROVE). Ready to merge after deploy + compat 52/52 on the merge head.

Issue #371, opened 2026-07-07
server exchange-phase send failures ride the raw io catch-all — GT maps them to IESENDMESSAGE/IESENDRESULTS


therealevanhenry:
Found by PR #370's r1 (pre-existing on main — the head and main binaries emit byte-identical lines; #370 fixed the *teardown* for this path, not the prose).

When a server-side exchange-phase **send** fails (peer RSTs the ctrl socket before/while the server writes its half), the error rides the serve-loop generic catch-all (`Err(e)` arm, server.rs:631) and prints the raw io Display:

```
riperf3: error - Connection reset by peer (os error 104)
```

GT 3.21 live-probed on the same scenario (ExchangeResults state-write against an RST'd ctrl):

```
iperf3: error - unable to send control message - port may not be available, the other side may have stopped running, etc.: Transport endpoint is not connected
```

i.e. **IESENDMESSAGE(111)** — the class #345 already maps for the pre-test `send_state` failure via `RiperfError::SendControlFailed`.

The exchange phase has three sends collapsing to the raw catch-all, mapping to two GT classes:

- `send_state(ExchangeResults)` and `send_state(DisplayResults)` → **IESENDMESSAGE** (`unable to send control message - port may not be available, the other side may have stopped running, etc.`)
- `send_results` → **IESENDRESULTS** (`unable to send results`)

Fix shape: route the exchange-phase sends through the #345 surfaces (typed variants + the stamped serve-loop arms + wire-back is N/A here since the ctrl is already dead — verify GT's cleanup_server behavior on this path when implementing). Suffix convention: honest errno-0 / live strerror per the #336/#345 precedent, not GT's stale-errno form.

House probes: r1's harness = mock peer completing through the results-exchange state write, then SO_LINGER(0) RST before the server's send. GT surface deterministic on this host.

therealevanhenry:
Addendum from PR #370's r2 (live-probed): the stderr prose is not the only gap on this class — the **-J doc shape** diverges too.

GT 3.21 `-s -1 -J` on the exchange-send failure emits a **populated** doc — `start`/`intervals`/`end` with streams, sums, cpu — plus the IESENDMESSAGE error key (json_finish runs over the reporter_callback output that already happened at TEST_END). riperf3 emits the #330 skeleton (`end:{}` + error key/event), byte-identical on main and the #370 head.

[...]

Issue #267, opened 2026-06-23
Failed-exchange stderr wording diverges from GT; SIGKILLed server should emit empty end block


therealevanhenry:
> **GT** = ground truth: the pinned reference **iperf 3.21**, built from source at commit `d39cf41526626b4e5a130f115d931cd6cbdffc19`; source-line citations refer to that tree.

From the 0.7.4 cold-review ledger (noted in #246's issue).

## Problem

Two related divergences on the failed-results-exchange / abrupt-peer-loss path:

- **stderr wording:** GT prints `control socket has closed unexpectedly` where riperf3 prints `peer disconnected` for the same failure class.
- **SIGKILL'd-server shape:** when the server is `SIGKILL`'d mid-test, GT emits an empty `end{}` block; riperf3's shape differs.

## Fix direction

Match GT's stderr wording for the control-socket-closed class and emit the empty `end{}` on the abrupt-loss path. Map the exact GT message via `iperf_error.c` / `iperf_handle_message_client`. Sibling of the perr-suffix work already filed as #248.

Compat / faithfulness.

therealevanhenry:
GT probes (3.21 @ d39cf41, live SIGKILL of the server mid-test): TEXT client prints interval rows then exactly `iperf3: error - control socket has closed unexpectedly` (exit 1, NO summary block). `-J` client emits ONE doc: **fully-populated start** (all late fields — the test reached TestStart), **collected intervals** (2 for a 2 s window), **bare `end: {}`** (the errexit class never populates json_end), and `error: control socket has closed unexpectedly`. riperf3 today: Display = 'peer disconnected' and the CLI -J sink emits the EMPTY skeleton (start: connected/version/system_info only, no intervals) — so the fix is (a) the Display wording and (b) rendering the doc from run state on this path: exactly #281's bare_end=true + Started stage (full start + intervals + end:{}).

PR #407, opened 2026-07-09
feat: Client::run returns RunOutcome; flip library output default to quiet (#293, #294)

### LINKED ISSUE #293: Unify Client::run abnormal-end semantics: return the partial Report on server-terminate/relayed-error paths

### LINKED ISSUE #294: Flip the library default to quiet output
From the 0.8.0 post-release review (v0.7.4..v0.8.0, main @9e0f77e). **Breaking** — 0.9.0. Companion to #290 (the 0.8.1 additive output knob; depends on it landing first).

## Problem

Once the additive `.emit_output(bool)` knob exists (#290), the remaining design question is the *default*: should a bare `ClientBuilder::new(..).build()?.run()` write the full iperf3 text output to the host process's stdout?

For a library, printing-unless-asked-not-to is a magic default. The faithful-output contract belongs to the **CLI** surface — and the CLI constructs its builder explicitly and can (and should) set `emit_output(true)` itself, so flipping the library default touches only direct library consumers, none of the wire/CLI behavior.

## Proposal

In 0.9.0, flip the builder default to quiet: `run`/`run_once` return the `Report` and emit nothing unless `.emit_output(true)` is set. The CLI passes `true` explicitly; `-J`, `--json-stream`, text, `--get-server-output` capture — all unchanged for CLI users.

## Notes

- Breaking: observable behavior change for existing library callers who rely on the printed output → 0.9.0.
- Decide together with the abnormal-end outcome issue — both reshape the same "library-first `run`" contract and should ship in one migration.

therealevanhenry:
Closes #293. Closes #294. The 0.9.0 library-first `run` contract migration (client side).

Both issues reshape the same "library-first `run`" surface, so they ship together.

## #293 — `Client::run` returns `RunOutcome`

`Client::run` now returns `Result<RunOutcome>`, where `RunOutcome { report, termination }` and `Termination` is `Completed | Interrupted | ServerTerminated | ServerError(String)`. The rule:

- **`Ok(RunOutcome)`** for every run that produced a report — the measured `Report` plus how it ended.
- **`Err`** for runs that produced no report (failed connect / control handshake). Two rarer classes — `ControlSocketClosed` (#267) and `RecvResultsFailed` (#374) — still return `Err` despite emitting a doc; folding them in is a follow-up (noted in the module docs).

This fixes both problems the issue names:
1. **Lossy** — a server-terminated or relayed-`SERVER_ERROR` run used to build + print the partial report and then discard it behind an `Err`; a library caller couldn't reach it. Now it's `Ok(RunOutcome)` carrying that report.
2. **Inconsistent** — "ended early with partial data" was signaled two ways (interrupt = `Ok(Report{error})`, server-terminate = `Err`). Now every such ending is `Ok(RunOutcome)` with the ending in `termination`, and the compiler forces callers to handle each case.

The `RiperfError::ServerTerminated` / `ServerErrorRelayed` variants are **removed** (they're `Termination`s now).

### Faithfulness / scope

This changes **only the Rust return shape** — the wire bytes, the text/JSON the CLI prints, and the process exit codes are byte-identical. The CLI reads `termination` and maps it to the exit code it already produced: `Completed`/`Interrupted` → 0, `ServerTerminated`/`ServerError` → 1 (via `AbnormalExit`, which carries the errexit message so the `riperf3: error - <msg>` line renders exactly as before; the JSON modes suppress a second render since the lib already emitted the doc). iperf3 has no analogous inconsistency — it returns a bare `int` and prints in-band — so this is a pure library-API-ergonomics improvement, not a divergence.

**`Server::run_once` keeps its `Report` return in this PR.** #293 notes it "should get the same treatment for symmetry," but the server's termination taxonomy is much richer (client-terminated, control-closed, the #371 send classes, self-terminate, unknown-message) and needs its own design — filed as a follow-up, stacked on this PR (task/issue to follow).

## #294 — quiet library default

The builder `emit_output` default flips to `false`: a bare `run()`/`run_once()` returns the report and prints nothing. The faithful-output contract belongs to the **CLI**, which now sets `.emit_output(true)` explicitly. `-J`, `--json-stream`, text, `--get-server-output` — all unchanged for CLI users.

## Tests + verification

- Rewrote the ~6 server-terminate / relayed-error tests to the new contract (`Ok` + `termination` match) and swept the ~30 report-field call sites (`.report.end` etc.); added a `Completed`-on-clean pin and a pin for the previously-untested **any-state** `ServerTerminate` arm (a pre-existing gap the mutation check exposed).
- Mutation-checked the key mappings: mid-test + any-state `ServerTerminated` arms and the `errexit_message` mapping all red their pins.
- Green: 414 lib tests, all integration / quiet_output / bound_server / stream / teardown suites, full CLI suite (bar the known #349-family load flakes, green isolated). `fmt` + `clippy -D warnings` clean.

## Migration (for library consumers)

```
let report = client.run().await?;            // before
let report = client.run().await?.report;     // after (or match .termination)
```

Bare library runs are now silent; add `.emit_output(true)` to restore iperf3's printed output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,19 @@ This changelog begins at 0.6.0. For earlier releases (0.1.1–0.5.4), see the
 [git history](https://github.com/therealevanhenry/riperf3/commits/main) and
 release tags.
 
-## [0.8.0] - 2026-06-28
+## [Unreleased]
+
+### Breaking
+
+- `Client::run` returns `RunOutcome { report, termination }` instead of `Report` (#293).
+  A server-terminated or relayed-error run is now `Ok` — carrying the partial report plus
+  `Termination::ServerTerminated`/`ServerError(msg)` — instead of an `Err` that discarded it;
+  `Err` is reserved for runs that produced no report. The `RiperfError::ServerTerminated` /
+  `ServerErrorRelayed` variants are removed. Migration: `client.run().await?` →
+  `client.run().await?.report`; branch on `outcome.termination` for the ending.
+- The builder output default flips to quiet (#294): a bare `run()`/`run_once()` returns the
+  report and prints nothing. Opt into iperf3's full text/JSON output with `.emit_output(true)`
+  (the CLI sets this). Wire protocol, CLI output, and exit codes are unchanged.
 
 Architecture-and-API release. Wire protocol and CLI flags unchanged; success-path
 `-J`/text output byte-identical. Breaking changes are library-API only.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,13 +17,17 @@ release tags.
 
 - `Client::run` returns `RunOutcome { report, termination }` instead of `Report` (#293).
   A server-terminated or relayed-error run is now `Ok` — carrying the partial report plus
-  `Termination::ServerTerminated`/`ServerError(msg)` — instead of an `Err` that discarded it;
-  `Err` is reserved for runs that produced no report. The `RiperfError::ServerTerminated` /
+  `Termination::ServerTerminated`/`ServerError(msg)` — instead of an `Err` that discarded it.
+  `Err` now covers runs that produced no report (connect/handshake failures; the
+  `ControlSocketClosed`/`RecvResultsFailed` classes stay `Err` for now — a follow-up).
+  The `RiperfError::ServerTerminated` /
   `ServerErrorRelayed` variants are removed. Migration: `client.run().await?` →
   `client.run().await?.report`; branch on `outcome.termination` for the ending.
 - The builder output default flips to quiet (#294): a bare `run()`/`run_once()` returns the
   report and prints nothing. Opt into iperf3's full text/JSON output with `.emit_output(true)`
   (the CLI sets this). Wire protocol, CLI output, and exit codes are unchanged.
+
+## [0.8.0] - 2026-06-28
 
 Architecture-and-API release. Wire protocol and CLI flags unchanged; success-path
 `-J`/text output byte-identical. Breaking changes are library-API only.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -793,7 +793,7 @@ dependencies = [
 
 [[package]]
 name = "riperf3"
-version = "0.8.0"
+version = "0.9.0"
 dependencies = [
  "base64",
  "libc",
@@ -815,7 +815,7 @@ dependencies = [
 
 [[package]]
 name = "riperf3-cli"
-version = "0.8.0"
+version = "0.9.0"
 dependencies = [
  "clap",
  "libc",
@@ -831,7 +831,7 @@ dependencies = [
 
 [[package]]
 name = "riperf3-test-support"
-version = "0.8.0"
+version = "0.9.0"
 dependencies = [
  "serde_json",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.8.0"
+version = "0.9.0"
 authors = ["Evan Henry"]
 edition = "2021"
 # MSRV floor 1.85 (review r1, verified empirically): the locked base64ct 1.8
--- a/riperf3-cli/Cargo.toml
+++ b/riperf3-cli/Cargo.toml
@@ -14,7 +14,7 @@ readme = "../README.md"
 
 [dependencies]
 # riperf3 workspace dependencies
-riperf3 = { path = "../riperf3", version = "0.8.0" }
+riperf3 = { path = "../riperf3", version = "0.9.0" }
 
 # external dependencies
 clap.workspace = true
--- a/riperf3-cli/src/cli.rs
+++ b/riperf3-cli/src/cli.rs
@@ -573,7 +573,10 @@ impl Cli {
             return Err(END_CONDITIONS_MSG.into());
         }
 
-        let mut builder = riperf3::ClientBuilder::new(host);
+        // #294: the CLI is the faithful-output surface, so it opts INTO the
+        // full iperf3 text/JSON output explicitly. The library default is now
+        // quiet (a bare `run()` returns the Report and prints nothing).
+        let mut builder = riperf3::ClientBuilder::new(host).emit_output(true);
 
         // Format (#241): case-sensitive [kmgtKMGT], uppercase = byte-rates;
         // absent → the library's adaptive default ('a'), like iperf3 (#221).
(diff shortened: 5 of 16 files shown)


therealevanhenry:
**Review round 1 (cold adversarial agent, worktree @13f8fcf): REQUEST-CHANGES → all fixed @dd0df2a.**

The reviewer built the worktree and live-probed every abnormal ending against a GT server, confirming the load-bearing promise holds: CLI output + exit codes are **byte-for-byte and exit-code identical to main** and match GT (modulo program name) across the server-error refusal (text: both `SERVER ERROR -` and `error -` lines, no dup, exit 1; -J: single doc + error key, silent stderr), the server-terminate (text: partial summary + `error - the server has terminated`, exit 1; -J: single doc), and a local SIGINT (exit 0 + interrupt notice). drain-once verified (text→partial_report build-only, JSON→emit_results, mutually exclusive, each drains once; text PRINT output unchanged). AbnormalExit errexit strings match the removed Display strings exactly; derives + non_exhaustive correct; #294 coverage complete (all 22 server expecteds + expected_client carry emit_output(true), none missed/doubled); tests bite.

Findings, all addressed:
- **F1 (must-fix)**: my CHANGELOG edit had accidentally DELETED the `## [0.8.0] - 2026-06-28` header, folding released-0.8.0 content under `[Unreleased]` (the release flow would have relabeled it 0.9.0 and erased the record). **Fixed @dd0df2a** — header re-inserted; the header progression is clean again (`[Unreleased]` → `[0.8.0]` → `[0.7.4]` …).
- **F2**: `is_clean_exit` was dead pub API with a false doc ("the CLI uses this" — it uses `errexit_message`). **Dropped** (errexit_message subsumes the clean check via `None`).
- **F3**: the module doc claimed "`Err` is reserved for runs that produced NO report" — but ControlSocketClosed (#267) and RecvResultsFailed (#374) still emit a doc then return Err. Defensible scope boundary (CLI faithful via the suppress-list), but the absolute was wrong. **Doc corrected** to name them as deferred exceptions (module doc + CHANGELOG + PR body).
- **F4**: stale cli.rs comment ("bare ServerBuilder::new()" — no longer bare post-#294). **Updated.**
- **F5 (not a defect)**: a `wire_shape` RST-timing flake under full-suite load — pre-existing, unmodified path, disclosed.

All fixes are docs/changelog/dead-code only — no behavior change. Two cold rounds still to complete before this holds for your merge go.

therealevanhenry:
**Review round 2 (fresh cold adversarial agent, worktree @dd0df2a): APPROVE, no blockers.**

Independently re-derived (not trusting r1). Empirically re-probed all three abnormal endings with the worktree `riperf3 0.9.0` binary vs GT `iperf 3.21` and vs a `main`-built 0.8.0 baseline — no exit-code, stderr-line, duplicate-line, or JSON-document-count regression in any cell (server-error refusal text/-J exit 1; server-terminate text/-J exit 1; local SIGINT text/-J exit 0 with the interrupt notice, not an `error -` line). Worktree == main baseline byte-for-byte on the deterministic refusal.

All four r1 fixes verified correct:
- F1: CHANGELOG diff vs main is ONLY the `[Unreleased]` block; `[0.8.0] - 2026-06-28` restored with its full body intact; header sequence clean.
- F2: `is_clean_exit` fully gone (0 grep hits); the CLI uses `errexit_message`.
- F3: the Err-scope doc now names ControlSocketClosed/RecvResultsFailed as the deferred exceptions (verified both still `Err`, both in the CLI suppress-list).
- F4: the cli comment no longer says "bare".

Version bump: workspace + dep pin + Cargo.lock all 0.9.0; `--version` derives from the package (no hardcoded string); `cargo semver-checks --baseline-rev main` classifies 0.8.0→0.9.0 as the (0.x) major-equivalent that permits the break — passes. drain-once clean across every terminal path (the new text-mode `partial_report` drain replaces a path that didn't drain on main). errexit strings byte-exact to the removed Displays; `perr_tail` not appended to AbnormalExit; JSON single-doc confirmed. Tests strengthened, not weakened (every expect_err→expect asserts the specific Termination). Full suite 0 failed; clippy clean.

Non-blocking:
- **N1 (closed @b971fec)**: no test asserted `Termination::Interrupted` on a local signal — and since Interrupted and Completed both exit 0, a mis-map would slip past. Strengthened `interrupt_honored_in_setup_phase_wait` to assert `termination == Interrupted`; mutation-verified (Interrupted→Completed reds it).
- **N2 (out of scope)**: a stale `json_report.rs:434` comment implying 0.9.0 Option-ifies some fields — pre-existing on main, `json_report.rs` untouched here; noted for the release-gate docs scrub.
- Reminder: CHANGELOG stays `[Unreleased]` until the release step stamps `[0.9.0] - <date>` (normal pre-release state).

Two cold rounds complete (r1 RC → r2 APPROVE). Ready for deploy + compat 52/52, then Evan's merge go.

